PR #127, opened 2023-07-27
[DO NOT MERGE] Feat/glp and mock x rate oracles

avniculae:


products/dated-irs/cannonfiles/mock_x/arbitrum.toml
```diff
@@ -0,0 +1,10 @@
+# This cannonfile is meant to be run on Arbitrum testnets.
+include = [
+  "utils/commons.toml",
+  "utils/aave-rate-oracles.toml",
+  "utils/glp-rate-oracle.toml",
+]
+
+[setting.xChainId]
```
ioanaroceanu2000, line 8: what does x stand for?

products/dated-irs/src/mocks/IMockXRateOracle.sol
```diff
@@ -0,0 +1,18 @@
...
+may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://github.com/Voltz-Protocol/v2-core/blob/main/products/dated-irs/LICENSE
+*/
+pragma solidity >=0.8.19;
+
+import { UD60x18 } from "@prb/math/UD60x18.sol";
+
+interface IMockXRateOracle {
```
ioanaroceanu2000, line 12: why do we need mocks?

products/dated-irs/src/oracles/GlpRateOracle.sol
```diff
@@ -0,0 +1,147 @@
...
+import {IRateOracle, IERC165} from "../interfaces/IRateOracle.sol";
+import {IRewardTracker} from "../interfaces/external/glp/IRewardTracker.sol";
+import {IVault} from "../interfaces/external/glp/IVault.sol";
+import {IRewardRouter} from "../interfaces/external/glp/IRewardRouter.sol";
+import {IGlpManager} from "../interfaces/external/glp/IGlpManager.sol";
+import { UD60x18, ud, mulDiv } from "@prb/math/UD60x18.sol";
+import {IERC20} from "@voltz-protocol/util-contracts/src/interfaces/IERC20.sol";
+
+/// @notice GP stands for GLP PRECISION (1e30)
+contract GlpRateOracle is IRateOracle {
```
ioanaroceanu2000, line 19: we used to have a function "refresh contracts" in case any of the GLp addresses change. do you think that's not necessary?

products/dated-irs/src/oracles/GlpRateOracle.sol
```diff
@@ -0,0 +1,147 @@
...
+    uint256 rewardsRateSinceLastUpdateGP = mulDiv(
+      cumulativeRewardPerTokenGP - state.lastCumulativeRewardPerTokenGP,
+      state.lastEthPriceInGlpGP,
+      GLP_PRECISION
+    );
+
+    UD60x18 rewardsRateSinceLastUpdate = ud(rewardsRateSinceLastUpdateGP).div(ud(GLP_PRECISION));
+
+    // compute index using rate increase & last index
+    liquidityIndex = state.lastIndex.add(rewardsRateSinceLastUpdate);
```
ioanaroceanu2000, line 138: just realised, have we thought how the accounting logic work with GLP... essentially the settlement CF is 
`base * maturityIndex + quote = notional/indexAtTrade * maturityIndex + notional * (loackedRate * time + 1)`
meaning `maturityIndex/indexAtTrade` should be the accrued variable interest?

products/dated-irs/cannonfiles/mock_x/arbitrum.toml
avniculae: cross

products/dated-irs/src/mocks/IMockXRateOracle.sol
avniculae: we need a way to push data from mainnet to testnet, and these mock contracts will allow for that

products/dated-irs/src/oracles/GlpRateOracle.sol
avniculae: I think this is not necessary as we can deploy a new contract in case glp contracts change (since I assume this doesn't happen too often right? have we ever used the refresh function before?)

products/dated-irs/src/oracles/GlpRateOracle.sol
avniculae: which is not the case since the glp rate oracle is linear rather then compounding, right?

PR #130, opened 2023-08-01
Feat/all exposures

ioanaroceanu2000:
Changes
- refactored get exposure logic in Portfolio to accommodate for multiple collateral types
- store active collateral types in the portfolio (it doesn't seem necessary to deactivate them, the logic will be too complex for the small gas savings in the future)
- exposes a function that returns an array of trader/makerShort/makerLong exposures for all collateral types ever activated by the account


products/dated-irs/src/storage/ExposureHelpers.sol
```diff
@@ -0,0 +1,179 @@
+/*
```
ioanaroceanu2000, line 1: I created this helper library and moved all functions from the Portfolio that do not depend on the state
Maybe this should not stay in the `storage` folder

arturbeg (approved):

PR #139, opened 2023-08-04
Feat/account refactoring

0xZenus:
- dropped liquidator booster logic
- merged **market fee configuration** and **market risk configuration** into **market**
- added account trackers (e.g. collateral balance, active collaterals, etc.)
- added collateral pool trackers (e.g. collateral balance)
- kicked off initiative of tracking all storage data with events (for clean off-chain picture) and pushed comments with todo's  (just for core for now, will need to do the same for the product and exchange storages as well)
- added MarginRequirements struct to pass all (im, lm, highestULoss, isImSatisfied, isLmSatisfied) easier. 
- replaced **load** with **exists** for markets and accounts
- got rid of MarketCreator and added create function to Market to isolate CollateralPool dependency more
- added collateral configuration utilities (such as (weighted) USD -> token, token -> (weighted) USD convertions)
- added helper libraries for Account to split functionalities


core/src/libraries/AccountCollateral.sol
```diff
@@ -0,0 +1,168 @@
...
+
+            // check if there's any available balance in USD
+            if (weightedBalanceInUSD >= mrInUSD.initialMarginRequirement + mrInUSD.highestUnrealizedLoss) {
+                // get the available weighted balance in USD
+                uint256 availableWeightedBalanceInUSD = 
+                    weightedBalanceInUSD - mrInUSD.initialMarginRequirement - mrInUSD.highestUnrealizedLoss;
+
+                // convert weighted balance in USD to collateral
+                uint256 availableAmountInCollateral = 
+                    CollateralConfiguration.load(collateralType).getWeightedUSDInCollateral(availableWeightedBalanceInUSD);
```
arturbeg, line 142: hmm, does weighting need to be applied in here to get the amount in collateral terms?

core/src/storage/Account.sol
```diff
@@ -344,232 +181,123 @@ library Account {
...
-     * @dev Returns the account stored at the specified account id.
+     * @dev Changes the account mode.
      */
-    function load(uint128 id) internal pure returns (Data storage account) {
-        require(id != 0);
-        bytes32 s = keccak256(abi.encode("xyz.voltz.Account", id));
-        assembly {
-            account.slot := s
+    function changeAccountMode(Data storage self, bool isMultiToken) internal {
+        if (self.isMultiToken == isMultiToken) {
```
arturbeg, line 263: hmm, does this function need an input vs. just flipping the mode?

core/src/storage/ProtocolRiskConfiguration.sol
```diff
@@ -13,6 +13,13 @@ import {UD60x18} from "@prb/math/UD60x18.sol";
  * @title Tracks protocol-wide risk settings
  */
 library ProtocolRiskConfiguration {
+    /**
+     * @notice Emitted when the protocol risk configuration is created or updated
+     * @param config The object with the newly configured details.
+     * @param blockTimestamp The current block timestamp.
+     */
+    event ProtocolRiskConfigurationUpdated(Data config, uint256 blockTimestamp);
```
arturbeg, line 21: this is now essentially CollateralPoolRiskConfiguration?

arturbeg (approved):

core/src/libraries/AccountCollateral.sol
0xZenus: My thinking is the following: Let's consider a multi token account with one collateral A, balance 100 and weight 0.9 and price 1. Your weighted balance in USD is 90. Let's consider that the IMR of the account is 45, which leaves 45 free weighted amount. However, you need to remove the weight when getting free amount in collateral, because the account can freely withdraw 50 collateral A. Let me know if I miss anything.

core/src/storage/Account.sol
0xZenus: I actually wanted to check with another fact. What if we convert account mode to Enum ( with single-token and multi-token for now) but it gives us flexibility of adding more modes. And at this point, changeMode will have input newMode. What do you think?

core/src/storage/ProtocolRiskConfiguration.sol
0xZenus: Yes, we'll do this in a separate task. We need to decide what params will be protocol level or collateral pool level.

core/src/storage/Account.sol
arturbeg: oh yeah, good point, enum is a good idea since we'll most likely need new modes

core/src/storage/ProtocolRiskConfiguration.sol
arturbeg: okay, sounds good

core/src/libraries/AccountCollateral.sol
arturbeg: oh i see, that makes sense

PR #140, opened 2023-08-07
Feat/further separation

0xZenus:


core/src/interfaces/ICollateralConfigurationModule.sol
```diff
@@ -25,7 +25,7 @@ interface ICollateralConfigurationModule {
      * Emits a {CollateralConfigured} event.
      *
      */
-    function configureCollateral(CollateralConfiguration.Data memory config) external;
+    function configureCollateral(address tokenAddress, CollateralConfiguration.Config memory config) external;
```
arturbeg, line 28: why's the token address separate?

core/src/libraries/AccountCollateral.sol
```diff
@@ -123,9 +121,23 @@ library AccountCollateral {
...
+                // return the available collateral balance
+                collateralBalanceAvailable = collateralBalance - mr.initialMarginRequirement - mr.highestUnrealizedLoss;
+            }
+        }
+
+        if (self.accountMode == Account.MULTI_TOKEN_MODE) {
             // get im and lm requirements and highest unrealized pnl in USD
-            AccountExposure.MarginRequirements memory mrInUSD = 
+            Account.MarginRequirement memory mrInUSD = 
                 self.getMarginRequirementsAndHighestUnrealizedLoss(collateralType);
```
arturbeg, line 141: this function getMarginRequirementsAndHighestUnrealizedLoss already deals with the fact wether the account is multi-token or single-token by relying on the state, would it not be cleaner to make it more stateless by having the mode an input?

arturbeg: my feeling is it may make testing easier but keen to hear your thoughts

0xZenus: Let me put a ticket to discuss a bit later about this. However, we can do this in another PR.

core/src/interfaces/ICollateralConfigurationModule.sol
0xZenus: just because I de-coupled config and cachedConfig. Now you're saying: set config for this particular collateral address, rather than appending collateral address in the struct.

core/src/libraries/AccountCollateral.sol
arturbeg: sounds good

core/src/interfaces/ICollateralConfigurationModule.sol
arturbeg: is cached config immutable?

0xZenus: yes. you can't modify it.

PR #141, opened 2023-08-07
Collateral pool-free accounts and collateral pools

0xZenus:


core/src/interfaces/IAccountModule.sol
```diff
@@ -115,7 +71,7 @@ interface IAccountModule {
      *
      * Emits a {AccountCreated} event.
      */
-    function createAccount(uint128 requestedAccountId, address accountOwner, uint8 accountMode) external;
+    function createAccount(uint128 requestedAccountId, address accountOwner, bytes32 accountMode) external;
```
arturbeg, line 74: what's the motivation behind using bytes32 instead of uint8 for the account mode?

core/src/libraries/AccountCollateral.sol
```diff
@@ -59,10 +59,10 @@ library AccountCollateral {
             }
         }
 
-        // update the corresponding collateral pool balance
-        Market.exists(self.firstMarketId)
-            .getCollateralPool()
-            .increaseCollateralBalance(collateralType, amount);
+        // update the corresponding collateral pool balance if exists
+        if (self.firstMarketId > 0) {
```
arturbeg, line 63: !=0 instead of > 0 to be consistent with other places that use ==0?

core/src/libraries/AccountRBAC.sol
```diff
@@ -0,0 +1,166 @@
...
+import "@voltz-protocol/util-contracts/src/helpers/SetUtil.sol";
+import "@voltz-protocol/util-contracts/src/errors/AddressError.sol";
+
+import "../storage/Account.sol";
+
+/**
+ * @title Object for tracking an accounts permissions (role based access control).
+ */
+
+library AccountRBAC {
```
arturbeg, line 19: what's the motivation behind having rbac as fully stateless?

core/src/modules/FeeConfigurationModule.sol
```diff
@@ -12,25 +12,47 @@ import "../interfaces/IFeeConfigurationModule.sol";
...
     }
 
     /**
      * @inheritdoc IFeeConfigurationModule
      */
-    function getMarketFeeConfiguration(uint128 marketId)
+    function configureCollateralPoolMarketFee(uint128 marketId, Market.FeeConfiguration memory config) external override {
+        Market.Data storage market = Market.exists(marketId);
+        market.getCollateralPool().onlyOwner();
+        market.setProtocolFeeConfiguration(config);
```
arturbeg, line 32: setCollateralPoolFeeConfiguration?

core/src/modules/MarketManagerModule.sol
```diff
@@ -110,17 +108,26 @@ contract MarketManagerModule is IMarketManagerModule {
...
-        fee = distributeFees(
+        uint256 protocolFee = distributeFees(
+            accountId, 
+            market.protocolFeeConfig.feeCollectorAccountId, 
+            market.protocolFeeConfig.atomicTakerFee, 
+            collateralType, 
+            annualizedNotional
+        );
+
+        uint256 collateralPoolFee = distributeFees(
```
arturbeg, line 119: hmm, can we not call distribute fees ones for both?

core/src/storage/Account.sol
```diff
@@ -119,29 +160,38 @@ library Account {
...
+        // disallowing account ID 0 means we can use a non-zero accountId as an existence flag in structs like Position
+        if (id == 0) {
+            revert AccountAlreadyExists(id);
+        }
 
+        // load the account data
         account = load(id);
 
+        // if the account id is non-zero, it means that the account has already been created
+        if (account.id > 0) {
```
arturbeg, line 177: !=0?

arturbeg: same result just feels cleaner

core/src/storage/CollateralPool.sol
```diff
@@ -94,33 +140,53 @@ library CollateralPool {
...
+        self.merge(child);
+    }
+
+    function revokeMergeProposal(Data storage self) internal {
+        self.proposedParentId = 0;
+    }
 
-        self.rootId = root.id;
-        emit CollateralPoolUpdated(self.id, self.rootId);
+    function merge(Data storage parent, Data storage child) internal {
```
arturbeg, line 165: will be good to talk in detail through this root logic

core/src/interfaces/IAccountModule.sol
0xZenus: Human-readable values (i.e. "SINGLE-TOKEN" and "MULTI-TOKEN" vs. 1 and 2)

core/src/libraries/AccountCollateral.sol
0xZenus: 51e0786b9daf796728752acc8bbb5def6155f8f4

core/src/libraries/AccountRBAC.sol
0xZenus: it was used only by account. I couldn't track the updates when the library was stateful and generic. E.g. grantPermission emits an event but we don't know where to map to.

core/src/storage/CollateralPool.sol
0xZenus: sure. we can talk through it.

core/src/modules/FeeConfigurationModule.sol
0xZenus: Great catch! 300a8b756e8e449c7d8f2ae2b958c43c72fd3e67

arturbeg (approved):

PR #146, opened 2023-08-17
[WIP] feat: insurance fund config

ioanaroceanu2000:


core/src/storage/Account.sol
```diff
@@ -279,6 +279,9 @@ library Account {
     }
 
     function increaseCollateralBalance(Data storage self, address collateralType, uint256 amount) internal {
```
ioanaroceanu2000, line 281: unrelated but felt the need to add this check

core/src/modules/AutoExchangeModule.sol
```diff
@@ -58,13 +59,16 @@ contract AutoExchangeModule is IAutoExchangeModule {
...
-
-        bool _isEligibleForAutoExchange = 
-            account.isEligibleForAutoExchange(quoteType);
+        CollateralPool.InsuranceFundConfig memory insuranceFundConfig = 
+            Market.load(account.firstMarketId).getCollateralPool().insuranceFundConfig;
+        Account.Data storage insuranceFundAccount = Account.exists(
+            insuranceFundConfig.accountId
+        );
 
+        bool _isEligibleForAutoExchange = account.isEligibleForAutoExchange(quoteType);
```
0xZenus, line 71: I'd suggest to get rid of the underline for variable names. It decreases the readability in general.

ioanaroceanu2000: foundry was giving a warning because the function name is the same as the variable name. are you proposing a different variable name?

0xZenus: yes, we can use isAccountEligibleForAutoExchange or similar.

ioanaroceanu2000: fixed [1078ab6](https://github.com/Voltz-Protocol/v2-core/pull/146/commits/1078ab67941dc0ae57172ed9efd6ca13ad82d819)

0xZenus (approved):

PR #147, opened 2023-08-18
Re-review insurance fund

0xZenus:


core/src/modules/FeeConfigurationModule.sol
```diff
@@ -44,6 +44,21 @@ contract FeeConfigurationModule is IFeeConfigurationModule {
...
 
+    /**
+     * @inheritdoc IFeeConfigurationModule
+     */
+    function configureCollateralPoolInsuranceFundMarketFee(
+        uint128 marketId,
+        UD60x18 insuranceFundMakerFee,
+        UD60x18 insuranceFundTakerFee
+    ) external override {
+        OwnableStorage.onlyOwner();
```
0xZenus, line 55: This should be restricted to collateral pool owner, no protocol owner.

core/src/modules/MarketManagerModule.sol
```diff
@@ -155,42 +156,37 @@ contract MarketManagerModule is IMarketManagerModule {
...
-        // distribute a portion of collateralPoolFee into the insurance fund 
-        CollateralPool.InsuranceFundConfig memory insuranceFund = 
-            market.getCollateralPool().insuranceFundConfig;
-        distributeFees(
-            accountId,
-            insuranceFund.accountId, 
-            insuranceFund.makerAndTakerFee,
-            collateralType, 
-            annualizedNotional
+        Market.Data memory market = Market.exists(marketId);
```
0xZenus, line 180: why don't we pass marketId to the propagateOrder function and grab the Market.Data in there?

core/src/storage/Market.sol
```diff
@@ -93,6 +93,16 @@ library Market {
...
+        /**
+         * @dev Insurance Fund Maker Fee is multiplied by the annualised notional traded
+         * @dev to derived the maker fee deposited in the insurance fund.
+         */
+        UD60x18 insuranceFundMakerFee;
+        /**
+         * @dev Insurance Fund Taker Fee is multiplied by the annualised notional traded
+         * @dev to derived the taker fee deposited in the insurance fund.
+         */
+        UD60x18 insuranceFundTakerFee;
```
0xZenus, line 105: For me it feels that there's incosistency with the fee configurations, i.e. collector id for collateral pool defined in the market and collector id for insurance fund defined in collateral pool. Why don't we maintain the consistency and have this collector id per market?

ioanaroceanu2000: my understanding is that the insurance fund id is meant to stay the same for all markets in the collateral pool. This is because the insurance fund covers bad debt of accounts, not markets. I think having different insurance fund accounts for markets will make covering bad debt more difficult.

core/src/modules/MarketManagerModule.sol
ioanaroceanu2000: because we need the market object on the external function level to get the maker/taker fees i.e.`market.config.atomicFee`

core/src/interfaces/IFeeConfigurationModule.sol
```diff
@@ -48,6 +48,21 @@ interface IFeeConfigurationModule {
...
+    /**
+     * @notice Creates or updates the insurance fund fees applied on the market's maker and taker orders 
+     * @param insuranceFundMakerFee Percentage of a maker order distrubuted to insurance fund.
+     * @param insuranceFundTakerFee Percentage of a taker order distrubuted to insurance fund.
+     *
+     * Requirements:
+     *
+     * - `msg.sender` must be the owner of the collateral pool.
+     */
+    function configureCollateralPoolInsuranceFundMarketFee(
```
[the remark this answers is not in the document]
ioanaroceanu2000: renamed in [ddc37eb](https://github.com/Voltz-Protocol/v2-core/pull/147/commits/ddc37ebb2aea88aff3e6c45271fe5893c224da78)

core/src/modules/FeeConfigurationModule.sol
ioanaroceanu2000: ah thanks! fixed in [ddc37eb](https://github.com/Voltz-Protocol/v2-core/pull/147/commits/ddc37ebb2aea88aff3e6c45271fe5893c224da78)

PR #149, opened 2023-08-22
feat: optional owner for feature flag

0xZenus:


utils/modules/src/storage/FeatureFlag.sol
```diff
@@ -22,6 +26,22 @@ library FeatureFlag {
         }
     }
 
+    function setOwner(Data storage feature, address owner) internal {
+        feature.owner = owner;
+    }
+
+    function checkOwner(Data storage feature) internal view {
```
avniculae, line 33: i would have named this `onlyOwner`, to match ownable storage

utils/modules/src/interfaces/IFeatureFlagModule.sol
```diff
@@ -11,6 +11,13 @@ pragma solidity >=0.8.19;
  * If a feature flag is set and then removed forever, consider deleting the line mentioned above from the function's body.
  */
 interface IFeatureFlagModule {
+    /**
+     * @notice Emitted when the feature owner is changed.
+     * @param feature The bytes32 id of the feature.
+     * @param owner The new owner address of the feature.
+     */
+    event FeatureFlagOwnerUpdated(bytes32 indexed feature, address owner);
```
avniculae, line 19: where is this emitted?

avniculae (approved):

utils/modules/src/interfaces/IFeatureFlagModule.sol
0xZenus: https://github.com/Voltz-Protocol/v2-core/pull/149/commits/7acc11d44cde34a267282f40340ded29075cca73

PR #155, opened 2023-09-04
Feat/circuit breaker

ioanaroceanu2000:


core/src/storage/CollateralPool.sol
```diff
@@ -433,4 +499,34 @@ library CollateralPool {
...
+
+    /**
+     * @dev Updates the withdraw limit trackers and checks if limit was reached
+     */
+    function checkWithdrawLimits(Data storage self, uint256 amount, address collateralType) internal {
+        // reset tracker if window has expired
+        uint32 currentWindowNumber = 
+            uint32(block.timestamp) / self.withdrawLimitsConfig[collateralType].withdrawalWindowSize;
+
+        uint256 windowWithdrawalsCopy = self.withdrawLimitsTrackers[collateralType].windowWithdrawals;
```
0xZenus, line 511: "Copy" suffix not needed since the name doesn't collide with any other variable.

core/src/storage/CollateralPool.sol
```diff
@@ -433,4 +499,34 @@ library CollateralPool {
...
+
+        bool isNewWindow = currentWindowNumber > self.withdrawLimitsTrackers[collateralType].windowNumber;
+        if (isNewWindow) {
+            self.withdrawLimitsTrackers[collateralType].windowNumber = currentWindowNumber;
+            windowWithdrawalsCopy = 0;
+        }
+
+        // check withdraw limits against tvl
+        windowWithdrawalsCopy += amount;
+        if (
```
0xZenus, line 521: It seems that these checks are part of `CollateralConfiguration.exists(collateralType).checkCollateralWithdrawLimits(amount, isNewWindow)`. Do I miss anything?

core/src/storage/CollateralPool.sol
ioanaroceanu2000: removed in [410b36b](https://github.com/Voltz-Protocol/v2-core/pull/155/commits/410b36be48240a7131d6ace174a2622412871dcf)

core/src/storage/CollateralPool.sol
```diff
@@ -433,4 +499,34 @@ library CollateralPool {
             revert Unauthorized(msg.sender);
         }
     }
+
+    /**
+     * @dev Updates the withdraw limit trackers and checks if limit was reached
+     */
+    function checkWithdrawLimits(Data storage self, uint256 amount, address collateralType) internal {
```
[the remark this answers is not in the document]
ioanaroceanu2000: yes, I think this is actually better because it allows for changing the window size without any worries. fixed in [410b36b](https://github.com/Voltz-Protocol/v2-core/pull/155/commits/410b36be48240a7131d6ace174a2622412871dcf)

core/src/storage/GlobalCollateralConfiguration.sol
```diff
@@ -0,0 +1,211 @@
...
+    function load(address collateralType) private pure returns (Data storage globalCollateral) {
+        bytes32 s = keccak256(abi.encode("xyz.voltz.GlobalCollateralConfiguration", collateralType));
+        assembly {
+            globalCollateral.slot := s
+        }
+    }
+
+    /**
+     * @dev Updates the withdraw limit trackers and checks if limit was reached
+     */
```
ioanaroceanu2000, line 186: - change to amount
- check window globally, because 2 collateral pools can have different windows

0xZenus (approved):

PR #157, opened 2023-09-07
feat: add new accounting logic

avniculae:


core/src/interfaces/ICollateralModule.sol
```diff
@@ -14,22 +14,22 @@ pragma solidity >=0.8.19;
 interface ICollateralModule {
```
0xZenus, line 14: I'd suggest emptying in-progress modules until we freeze the storages. Two reasons in my head:
  - functions are constantly changing and there's an overhead in maintaining modules
  - we need to re-think what functions need to append in here.

Keen to hear your thoughts as well.

core/src/interfaces/ILiquidationModule.sol
```diff
@@ -20,8 +20,8 @@ interface ILiquidationModule {
      * @param collateralType The collateral type of the account that is being checked
```
0xZenus, line 20: Same comment as in ICollateralModule.sol

PR #158, opened 2023-09-07
Liquidations Initial PR

arturbeg:


core/src/storage/Account.sol
```diff
@@ -69,6 +73,31 @@ library Account {
      */
     error AccountNotFound(uint128 accountId);
 
+    /**
+     * @dev Thrown when attempting to execute a bid in an expired liquidation bid priority queue
+     */
+    error LiquidationBidPriorityQueueExpired(uint256 queueId, uint256 queueEndTimestamp);
```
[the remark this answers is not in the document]
arturbeg: introduce AccountLiquidations (to maintain the split by functionalities)

core/src/interfaces/ICollateralConfigurationModule.sol
```diff
@@ -7,6 +7,8 @@ https://github.com/Voltz-Protocol/v2-core/blob/main/core/LICENSE
 */
 pragma solidity >=0.8.19;
 
+import "../storage/CollateralConfiguration.sol";
```
avniculae, line 10: remove

core/src/libraries/LiquidationBidPriorityQueue.sol
```diff
@@ -0,0 +1,97 @@
...
+            i = j;
+        }
+        delete self.liquidationBidsMap[toReturn];
+    }
+
+
+    function enqueue(Heap storage self, uint256 rank, LiquidationBid memory liquidationBid) internal {
+        if (self.ranks.length == 0) {
+            // todo: why initialize with a zero?
+            self.ranks.push(0); // initialize
```
avniculae, line 75: exactly, why?

core/src/libraries/LiquidationBidPriorityQueue.sol
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        self.ranks.push(rank);
+        uint256 i = self.ranks.length - 1;
+
+        while (i > 1 && self.ranks[i / 2] > self.ranks[i]) {
```
avniculae, line 81: this implies a min heap, do we want a max heap instead?

core/src/libraries/LiquidationBidPriorityQueue.sol
```diff
@@ -0,0 +1,97 @@
...
+            (self.ranks[i / 2], self.ranks[i]) = (rank, self.ranks[i / 2]);
```
avniculae, line 82: i would keep self.ranks[i] instead of rank

core/src/libraries/LiquidationBidPriorityQueue.sol
```diff
@@ -0,0 +1,97 @@
...
+    function top(Heap storage self) internal view notEmpty(self) returns(uint256) {
+        return self.ranks[1];
+    }
+
+    function topBid(Heap storage self) internal view notEmpty(self) returns(LiquidationBid memory) {
+        return self.liquidationBidsMap[top(self)];
+    }
+
+    function dequeue(Heap storage self) internal notEmpty(self) {
+        require(self.ranks.length > 1);
```
avniculae, line 44: do we want to revert if length is 1?

core/src/libraries/LiquidationBidPriorityQueue.sol
avniculae: we assume a 1-indexed array I think, no?

avniculae (commented):
This is a partial review

PR #160, opened 2023-09-08
Feat/matched orders

ioanaroceanu2000:


core/src/libraries/account/AccountExposure.sol
```diff
@@ -218,4 +219,19 @@ library AccountExposure {
...
+    internal
+    view
+    returns (uint256 totalMarketExposure) {
+        Market.Data storage market = Market.exists(marketId);
+        Account.MakerMarketExposure[] memory makerExposures = 
+            market.getAccountTakerAndMakerExposures(self.id);
+
+        // Aggregate LMR and unrealized loss for all exposures
+        for (uint256 i = 0; i < makerExposures.length; i++) {
+            totalMarketExposure += SignedMath.abs(makerExposures[i].lower.annualizedNotional);
```
0xZenus, line 233: This is not what we want. Lower and maker exposures double-count the taker exposure. As this is used only in taker orders, we can restrict it to total filled exposure.

core/src/libraries/actions/MatchedOrders.sol
```diff
@@ -0,0 +1,51 @@
+/*
+Licensed under the Voltz v2 License (the "License"); you 
+may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://github.com/Voltz-Protocol/v2-core/blob/main/core/LICENSE
+*/
+pragma solidity >=0.8.19;
+
+import "../../interfaces/external/IMarketManager.sol";
```
0xZenus, line 10: please add on-demand dependency (i.e. `import {...} from "...";`)

core/src/libraries/actions/MatchedOrders.sol
```diff
@@ -0,0 +1,51 @@
...
+
+        // verify counterparty account & access
+        Account.Data storage counterPartyAccount = 
+            Account.loadAccountAndValidatePermission(counterPartyAccountId, Account.ADMIN_PERMISSION, msg.sender);
+        counterPartyAccount.ensureEnabledCollateralPool();
+        
+        // execute orders
+        initialCounterPartyMarketExposure = counterPartyAccount.getTotalAbsoluteMarketExposure(marketId);
+
+        (bytes memory result,) =
```
0xZenus, line 43: why do we encode both taker order outputs as one? why can't we return two different bytes?

core/src/libraries/actions/MatchedOrders.sol
```diff
@@ -0,0 +1,51 @@
...
+        bytes[] memory orderInputs;
+        assembly {
+            counterPartyAccountId := calldataload(inputs.offset)
+            orderInputs := calldataload(add(inputs.offset, 0x20))
+        }
+
+        // verify counterparty account & access
+        Account.Data storage counterPartyAccount = 
+            Account.loadAccountAndValidatePermission(counterPartyAccountId, Account.ADMIN_PERMISSION, msg.sender);
+        counterPartyAccount.ensureEnabledCollateralPool();
```
0xZenus, line 38: counter party account needs to be in the same collateral pool as the initiator account. no need to check it again.

core/src/modules/ExecutionModule.sol
```diff
@@ -183,56 +186,97 @@ contract ExecutionModule {
...
+                
+            return abi.encode(matchResult, fee, counterPartyFee, counterPartyMarginRequirements);
         } else {
-            revert InvalidCommandType(command);
+            revert InvalidCommandType(commandName);
         }
     }
 
-}
+    /// @notice utility function
```
0xZenus, line 272: no need for this comment. private modifier already shows it.

core/src/modules/ExecutionModule.sol
```diff
@@ -183,56 +186,97 @@ contract ExecutionModule {
...
-                marketManager.completeOrder(accountId, marketId, inputs);
+                marketManager.completeOrder(accountId, command.marketId, command.inputs);
 
             if (cashflowAmount > 0) {
                 account.increaseCollateralBalance(collateralType, cashflowAmount.toUint());
             } else {
                 account.decreaseCollateralBalance(collateralType, (-cashflowAmount).toUint());
             }
             return abi.encode(result);
+        } else if (commandName == V2_MATCHED_ORDER) {
```
0xZenus, line 232: The commands have become quite messy. I just have a feeling that we can organise them better.

core/src/libraries/actions/MatchedOrders.sol
avniculae: where do we check that the counterparty account is in the same collateral pool though?

core/src/modules/ExecutionModule.sol
ioanaroceanu2000: I feel the same. there's a ticket regarding batched actions contract size and I'll consider this while looking at this ticket

core/src/modules/ExecutionModule.sol
ioanaroceanu2000: removed

core/src/libraries/actions/MatchedOrders.sol
ioanaroceanu2000: yh I think that's cleaner, changed in [467fd99](https://github.com/Voltz-Protocol/v2-core/pull/160/commits/467fd9992fac22a747a177c539ef1c85561fec94)

core/src/libraries/actions/MatchedOrders.sol
ioanaroceanu2000: changed

core/src/libraries/account/AccountExposure.sol
ioanaroceanu2000: Can you please have a look over `getAccountAbsoluteExposure` in Portfolio at [467fd99](https://github.com/Voltz-Protocol/v2-core/pull/160/commits/467fd9992fac22a747a177c539ef1c85561fec94). It subtracts the taker exposure after the double counting

ioanaroceanu2000: let me know if you can think of a cleaner solution

core/src/libraries/account/AccountExposure.sol
```diff
@@ -12,6 +12,7 @@ import {CollateralConfiguration} from "../../storage/CollateralConfiguration.sol
 import {CollateralPool} from "../../storage/CollateralPool.sol";
 import {Market} from "../../storage/Market.sol";
 
+import {SignedMath} from "oz/utils/math/SignedMath.sol";
```
0xZenus, line 15: is it used?

ioanaroceanu2000: double import, removed in [1db9abf](https://github.com/Voltz-Protocol/v2-core/pull/160/commits/1db9abfa86bdc24703bc85821efc2504fdb28732)

0xZenus (approved):

PR #162, opened 2023-09-15
Feat/rebased liquidations pq

avniculae:


core/src/storage/Account.sol
```diff
@@ -370,41 +496,466 @@ library Account {
...
+            }
+            collateralPool.updateInsuranceFundUnderwritings(quoteToken, insuranceFundDebit);
+        }
+
+        // execute adl orders (bankruptcy price is calculated in the market manager)
+
+        uint256[] memory markets = self.activeMarketsPerQuoteToken[quoteToken].values();
+        for (uint256 j = 0; j < markets.length; j++) {
+            uint128 marketId = markets[j].to128();
+            Market.exists(marketId).executeADLOrder(
```
ioanaroceanu2000, line 952: I would add `if shortfall > 0 then execute ADL`

core/src/storage/Account.sol
```diff
@@ -370,41 +496,466 @@ library Account {
...
+        Market.Data storage market = Market.exists(marketId);
+
+        LiquidationBidPriorityQueues storage liquidationBidPriorityQueues =
+        self.liquidationBidPriorityQueuesPerBubble[market.quoteToken];
+
+        // revert if the account is above dutch margin requirement & the liquidation bid queue is not empty
+
+        if (
+            liquidationBidPriorityQueues.priorityQueues
+        [liquidationBidPriorityQueues.latestQueueId].ranks.length > 0 && self.isAboveDutch(address(0))) {
```
ioanaroceanu2000, line 860: format

core/src/storage/Account.sol
```diff
@@ -370,41 +496,466 @@ library Account {
...
+        Account.Data storage liquidatorAccount = Account.exists(liquidatorAccountId);
+
+        collateralPoolsCheck(self.getCollateralPool().id, liquidatorAccount);
+
+        Market.Data storage market = Market.exists(marketId);
+
+        LiquidationBidPriorityQueues storage liquidationBidPriorityQueues =
+        self.liquidationBidPriorityQueuesPerBubble[market.quoteToken];
+
+        // revert if the account is above dutch margin requirement & the liquidation bid queue is not empty
```
ioanaroceanu2000, line 856: why do we care about the len of the bid queue?

core/src/interfaces/ICollateralConfigurationModule.sol
```diff
@@ -7,6 +7,8 @@ https://github.com/Voltz-Protocol/v2-core/blob/main/core/LICENSE
 */
 pragma solidity >=0.8.19;
 
+import "../storage/CollateralConfiguration.sol";
```
0xZenus, line 10: why?

core/src/interfaces/ILiquidationModule.sol
```diff
@@ -7,9 +7,53 @@ https://github.com/Voltz-Protocol/v2-core/blob/main/core/LICENSE
...
+    /**
+     * @notice Get the initial and liquidation margin requirements and highest unrealized loss
+     * (if maker positions were to be filled) along with the flags that specify if the initial or liquidation margin
+     * requirements are satisfied.
+     * @param accountId The id of the account that is being checked
+     * @param collateralType The collateral type for which the margin requirements are checked,
+     * where the collateral type is the centre of a given collateral bubble
+     * @return Margin requirement information object
+     */
+    function getMarginInfoByBubble(uint128 accountId, address collateralType)
```
0xZenus, line 27: is it still by bubble if we've got rid of single-bubble?

core/src/libraries/account/AccountCollateral.sol
```diff
@@ -119,7 +119,13 @@ library AccountCollateral {
         }
 
         Account.MarginInfo memory marginInfoCollateral = 
-            self.getMarginInfoByCollateralType(collateralType, self.getCollateralPool().riskConfig.imMultiplier);
+            self.getMarginInfoByCollateralType(
+                collateralType,
```
0xZenus, line 123: why do we fetch the risk config here and pass them to the function rather than fetching the risk config directly in that function?

core/src/libraries/account/AccountExposure.sol
```diff
@@ -50,33 +61,61 @@ library AccountExposure {
...
+            quoteToken,
+            imMultiplier,
+            mmrMultiplier,
+            dutchMultiplier,
+            adlMultiplier
+        );
 
         address[] memory tokens = CollateralConfiguration.exists(collateralPoolId, quoteToken).childTokens.values();
 
+        // todo: why do we need to loop through the margin requirements of child tokens when only base tokens
```
0xZenus, line 105: because this function is recursive in the collateral tree

core/src/interfaces/ICollateralConfigurationModule.sol
arturbeg: removed

core/src/interfaces/ILiquidationModule.sol
arturbeg: when we are performing a liquidation, we're assuming that auto-exchange has already taken place, note this point from the docs: https://www.notion.so/Reya-Book-e3c86f8b4178472fabe3de5f6b49bc95?d=dcd823cf2909453780dc861f4417eadb&pvs=4#4e3a07b85ced4dbc9aff5218da8efdb9, hence when we perform liquidations in a given market with a given quote token we want to be able to calculate the rewards in terms of a given quote (as a function of delta lm in a given bubble) token that are then paid out as a reward

0xZenus: got it. Should it be renamed then? (getMarginInfoInCollateral or so)?

core/src/libraries/account/AccountAutoExchange.sol
```diff
@@ -7,9 +7,160 @@ https://github.com/Voltz-Protocol/v2-core/blob/main/core/LICENSE
...
+    ) internal view returns (bool) {
+        AutoExchangeConfiguration.Data memory autoExchangeConfig = AutoExchangeConfiguration.load();
+
+        CollateralPool.Data storage collateralPool = self.getCollateralPool();
+        uint128 collateralPoolId = collateralPool.id;
+
+        // Single auto-exchange threshold check
+        {
+            Account.MarginInfo memory marginInfo =
+                self.getMarginInfoByCollateralType(collateralType,
```
0xZenus, line 43: clean-up idea: pass riskConfig struct in the function.

core/src/libraries/account/AccountExposure.sol
```diff
@@ -50,33 +61,61 @@ library AccountExposure {
...
+            adlDelta: getExchangedQuantity(marginInfo.adlDelta, exchange.price, exchange.haircut)
         });
     }
 
     function computeMarginInfoByBubble(
         Account.Data storage account, 
         uint128 collateralPoolId, 
         address quoteToken, 
-        UD60x18 imMultiplier
+        UD60x18 imMultiplier,
```
0xZenus, line 85: pass collateral pool risk config struct instead of separate fields.

integration/src/utils/SetupProtocol.sol
```diff
@@ -163,16 +163,25 @@ contract SetupProtocol is BatchScript {
...
-      })
-    );
+    // todo implement once liquidations are completed
+    // configureProtocolRisk(
+    //   collateralPoolId,
+    //   CollateralPool.RiskConfiguration({
+    //     imMultiplier:imMultiplier,
+    //     mmrMultiplier: mmrMultiplier,
+    //     liquidatorRewardParameter: liquidatorRewardParameter,
+    //     liquidationBidPriorityQueueDurationInSeconds: liquidationBidPriorityQueueDurationInSeconds,
```
0xZenus, line 180: very long variable names

products/dated-irs/src/storage/Portfolio.sol
```diff
@@ -439,4 +439,38 @@ library Portfolio {
             block.timestamp
         );
     }
+
+    function hasUnfilledOrders(Data storage self) internal view returns (bool) {
+        Market.Data storage market = Market.exists(self.marketId);
+
+        for (uint256 i = 1; i <= self.activeMaturities.length(); i++) {
```
0xZenus, line 446: you can fetch items via self.activeMaturities.values() and then use it as an array rather than interacting with the set each iteration.

products/dated-irs/src/storage/Portfolio.sol
```diff
@@ -439,4 +439,38 @@ library Portfolio {
...
+            }
+        }
+
+        return false;
+    }
+
+    function closeAllUnfilledOrders(Data storage self) internal returns (int256 closedUnfilledBasePool) {
+        Market.Data storage market = Market.exists(self.marketId);
+
+        for (uint256 i = 1; i <= self.activeMaturities.length(); i++) {
```
0xZenus, line 466: same as above.

core/src/storage/CollateralPool.sol
```diff
@@ -365,21 +449,27 @@ library CollateralPool {
         self.checkRoot();
 
         self.riskConfig.imMultiplier = config.imMultiplier;
-        self.riskConfig.liquidatorRewardParameter = config.liquidatorRewardParameter;
+        self.riskConfig.mmrMultiplier = config.mmrMultiplier;
+        self.riskConfig.unfilledOrderLiquidationPenaltyParameter = config.unfilledOrderLiquidationPenaltyParameter;
```
0xZenus, line 453: very long variable names. I'd suggest adding to RiskConfiguration a Liquidation struct that encapsulate all of these.

core/src/storage/CollateralPool.sol
```diff
@@ -102,11 +103,54 @@ library CollateralPool {
          * where IM = imMultiplier * LM
          */
         UD60x18 imMultiplier;
+
+        /**
+         * @dev MMR Multiplier (maintenance margin requirement multiplier)
+         * is used to introduce a buffer before liquidations occur to allow for liquidation bid submissions
+         * where MMR = mmrMultiplier * LM
+         */
+        UD60x18 mmrMultiplier;
```
0xZenus, line 112: I'd suggest adding a RiskMultiplier struct and another Liquidation struct in here rather than individual fields.

core/src/storage/CollateralPool.sol
0xZenus: Doing this, you can also assign riskConfig.struct at once rather than assigning field by field.

core/src/storage/Account.sol
```diff
@@ -163,18 +198,21 @@ library Account {
...
         /**
          * @dev First market id that this account is active on
          */
         uint128 firstMarketId;
+        /**
+         * @dev Liquidation Bid Priority Queues associated with the account alongside latest timestamp & id per
+         * collateral bubble
+         * collateralBubbleQuoteTokenAddress -> LiquidationBidPriorityQueues
+         */
+        mapping(address => LiquidationBidPriorityQueues) liquidationBidPriorityQueuesPerBubble;
```
0xZenus, line 214: Are they still by bubble? I think they are per collateral or quote type right?

core/src/modules/LiquidationModule.sol
```diff
@@ -17,5 +22,159 @@ import {ILiquidationModule} from "../interfaces/ILiquidationModule.sol";
...
+                liquidationBidPriorityQueues.latestQueueId,
+                liquidationBidPriorityQueues.latestQueueEndTimestamp
+            );
+        }
+
+        // extract top ranked order
+
+        LiquidationBidPriorityQueue.LiquidationBid memory topRankedLiquidationBid = liquidationBidPriorityQueues
+        .priorityQueues[
+        liquidationBidPriorityQueues.latestQueueId
```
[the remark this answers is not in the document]
0xZenus: Same comment from me and I'd suggest trimming down the variable names. It's very hard to read/follow.

core/src/modules/LiquidationModule.sol
```diff
@@ -17,5 +22,159 @@ import {ILiquidationModule} from "../interfaces/ILiquidationModule.sol";
...
+        ].topBid();
+
+        (bool success, bytes memory reason) = address(this).call(abi.encodeWithSignature(
```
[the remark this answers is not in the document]
0xZenus: I agree. This feels very hacky.

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -0,0 +1,561 @@
...
+        uint128 marketId;
+        bytes inputs;
+    }
+
+
+    /**
+     * @dev Checks if the account is below the maintenance margin requirement
+     * and reverts if that's not the case (i.e. reverts if the mmr requirement is satisfied by the account)
+     */
+    function isBelowMMRCheck(Account.Data storage self, address collateralType) internal view returns
```
0xZenus, line 121: are the following checks used in more places in the code? if not, it feels cleaner to me to have this code in-line rather than in separate functions.

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -0,0 +1,561 @@
...
+        LiquidationBidPriorityQueue.LiquidationBid memory liquidationBid
+    ) internal returns (uint256) {
+        // implement
+        // note, the ranking function should revert if the liquidation bid is attempting to liquidate more exposure
+        // than the user has
+        // also note, the ranking function should revert if the liquidation bid is attempting to touch non-active markets
+        return 0;
+    }
+
+    function submitLiquidationBid(
```
0xZenus, line 263: such long variable names in here, it makes the code un-readable. I'd suggest of trimming them down (e.g. liquidationBidPriorityQueueDurationInSeconds -> liquidationBidsMaxDuration)

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -0,0 +1,561 @@
...
+
+        if (marketIdsLength > collateralPool.riskConfig.maxNumberOfOrdersInLiquidationBid) {
+            revert LiquidationBidOrdersOverflow(marketIdsLength,
+                collateralPool.riskConfig.maxNumberOfOrdersInLiquidationBid);
+        }
+
+        for (uint256 i = 0; i < marketIdsLength; i++) {
+            uint128 marketId = liquidationBid.marketIds[i];
+            Market.Data storage market = Market.exists(marketId);
+            if (market.quoteToken != liquidationBid.quoteToken) {
```
0xZenus, line 242: since the liquidationBid contains marketId (implicitly quoteToken), why does it also contain quoteToken? Why is not fetched as market(marketId).quoteToken?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -0,0 +1,561 @@
...
+import {LiquidationBidPriorityQueue} from "../LiquidationBidPriorityQueue.sol";
+import { UD60x18, mulUDxUint } from "@voltz-protocol/util-contracts/src/helpers/PrbMathHelper.sol";
+import "../../interfaces/external/IMarketManager.sol";
+import { SafeCastU256, SafeCastI256 } from "@voltz-protocol/util-contracts/src/helpers/SafeCast.sol";
+
+/**
+ * @title Object for managing account liquidation utilities
+*/
+library AccountLiquidation {
+    using Account for Account.Data;
```
0xZenus, line 32: please go through all the functions and mark the functions that are not used out of this library as private rather than internal (to mark that those functions are utilities and don't interact with external code).

core/src/interfaces/ILiquidationModule.sol
arturbeg: hmm, it's still by bubble though, the concept of bubbles is still present however now we just no longer have the option to explicitely keep the account single-bubble when it comes to margin requirement checks

core/src/modules/LiquidationModule.sol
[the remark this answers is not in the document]
arturbeg: it does, originally i had the reverse orders if you remember and they we discussed introducing a try/catch pattern and this is the only way to introduce it, do you have other alternatives / options in mind?

core/src/storage/CollateralPool.sol
arturbeg: aren't all of the risk configs effectively liquidation related? well maybe except for the imMultiplier perhaps, is it still worth having a separate struct for liquidation related params?

arturbeg: however, liqudiation + risk multiplier decoupling makes sense

core/src/libraries/account/AccountAutoExchange.sol
arturbeg: https://github.com/Voltz-Protocol/v2-core/pull/162/commits/72678d842f0ff43e57cf50398435327f596a052c

core/src/libraries/account/AccountExposure.sol
arturbeg: https://github.com/Voltz-Protocol/v2-core/pull/162/commits/72678d842f0ff43e57cf50398435327f596a052c

integration/src/utils/SetupProtocol.sol
arturbeg: https://github.com/Voltz-Protocol/v2-core/pull/162/commits/72678d842f0ff43e57cf50398435327f596a052c

core/src/storage/CollateralPool.sol
arturbeg: https://github.com/Voltz-Protocol/v2-core/pull/162/commits/72678d842f0ff43e57cf50398435327f596a052c

core/src/storage/Account.sol
arturbeg: each quote type / settlement token has its own queue  (usdc, eth, etc can have their own queue but stETH wouldn't)-> each bubble is uniquely identifiable by its quote token (centroid), however, i think you're thinking in terms of a generic bubble setup which is potentially making things more complex vs. what we need -

core/src/libraries/account/AccountLiquidation.sol
arturbeg: done

core/src/libraries/account/AccountLiquidation.sol
arturbeg: good question, so the liquidation can only have markets that share the same quote token, perhaps we can avoid the need for an additional tracker in the bid by fetching the first market id, treating that as the quote token and then validating that all the other markets have the same quote token?

arturbeg: would that be cleaner vs. just having an additional quote token for a bid?

core/src/libraries/account/AccountLiquidation.sol
arturbeg: agreed, https://github.com/Voltz-Protocol/v2-core/pull/162/commits/586010a0e7f99440872744ff1eb73bc41f92dcc9

core/src/libraries/account/AccountLiquidation.sol
arturbeg: done https://github.com/Voltz-Protocol/v2-core/pull/162/commits/76aa4dd9cdf4f856edec7142cb4d19dc3b0a05c6

products/dated-irs/src/storage/Portfolio.sol
arturbeg: what's the benefit of this vs. dealing with sets? is this more gas efficient if yes, why do we need sets in the first place (not sure i fully get this)

arturbeg: while at it i'm a bit confused why we don't have smth like self.activeMaturities[marketId].values(), why isn't the active maturities set a function of the market id?

arturbeg: https://github.com/Voltz-Protocol/v2-core/pull/162/commits/82daebe3090ed0c856a250ac5ba91beaf3d4450e here's the commit with the suggested fix in any case

products/dated-irs/src/storage/Portfolio.sol
arturbeg: https://github.com/Voltz-Protocol/v2-core/pull/162/commits/82daebe3090ed0c856a250ac5ba91beaf3d4450e

core/src/libraries/LiquidationBidPriorityQueue.sol
```diff
@@ -0,0 +1,98 @@
...
+        mapping(uint256 => LiquidationBid) liquidationBidsMap;
+    }
+
+    modifier notEmpty(Heap storage self) {
+        require(self.ranks.length > 1);
+        _;
+    }
+
+    function top(Heap storage self) internal view notEmpty(self) returns(uint256) {
+        return self.ranks[1];
```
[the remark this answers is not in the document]
0xZenus: priority queues are usually 1-indexed because it's simpler to implement. When going up in the tree, you always do [i/2] rather than [(i-1)/2].

arturbeg (approved):

PR #168, opened 2023-09-25
Auto-Exchange Adjustments

arturbeg:


core/src/libraries/account/AccountAutoExchange.sol
```diff
@@ -37,63 +73,76 @@ library AccountAutoExchange {
...
         if (sumOfNegativeAccountValuesInUSD > autoExchangeConfig.totalAutoExchangeThresholdInUSD) {
             return true;
         }
 
-        if (totalAccountValueInUSD < 0) {
-            // todo: decide on what to do when the account is liquidatable
+
+        // Get total account value in USD
+        int256 totalAccountValueInUSD = 0;
+        {
```
ioanaroceanu2000, line 143: remove {} and assign directly

core/src/libraries/account/AccountAutoExchange.sol
```diff
@@ -106,52 +155,135 @@ library AccountAutoExchange {
...
-    ) internal view returns (uint256 /* coveringAmount */, uint256 /* autoExchangedAmount */ ) {
+        CollateralPool.Data storage collateralPool,
+        address quoteToken,
+        uint256 amountToAutoExchangeQuote
+    ) private view returns (uint256 amountToAutoExchange) {
+
+        Account.MarginInfo memory marginInfo =
+        self.getMarginInfoByCollateralType(
+            quoteToken,
+            collateralPool.riskConfig.riskMultipliers
```
ioanaroceanu2000, line 168: can't we pass risk multipliers as parameter instead of the whole collateralPool?

core/src/libraries/account/AccountAutoExchange.sol
```diff
@@ -106,52 +155,135 @@ library AccountAutoExchange {
...
-            uint256 correspondingTokenCoveringAmount = divUintUD(coveringTokenAmount, price.mul(autoExchangeDiscount));
-            return (correspondingTokenCoveringAmount, amountToAutoExchange);
+
+        UD60x18 autoExchangeInsuranceFee = CollateralConfiguration.load(
+            collateralPool.id,
+            quoteToken
+        ).baseConfig.autoExchangeInsuranceFee;
+
+
+        if (fromUD60x18(autoExchangeInsuranceFee) != 0) {
```
ioanaroceanu2000, line 269: seems like from UD60x18 divides autoExchangeInsuranceFee by 1e18. concerned this results in all values being 0. we should use `unwrap` for this

core/src/libraries/actions/AutoExchange.sol
```diff
@@ -7,18 +7,108 @@ https://github.com/Voltz-Protocol/v2-core/blob/main/core/LICENSE
...
-        // todo
+
+        Account.Data storage account = Account.exists(accountId);
+
+        if (!account.isEligibleForAutoExchange(quoteType)) {
+            revert AccountNotEligibleForAutoExchange(accountId, quoteType);
+        }
+
+        if (!account.isWithinBubbleCoverageExhausted(quoteType, collateralType)) {
+            revert WithinBubbleCoverageNotExhausted(accountId, quoteType, collateralType);
```
ioanaroceanu2000, line 56: can you please explain this condition?

core/src/libraries/account/AccountAutoExchange.sol
arturbeg: https://github.com/Voltz-Protocol/v2-core/pull/168/commits/83971b72731e8e4b81f4c1ee9903869108f62986

core/src/libraries/account/AccountAutoExchange.sol
avniculae: indeed. we should use either unwrap or intoUint256

arturbeg: good catch https://github.com/Voltz-Protocol/v2-core/pull/168/commits/f0311331ad5f08c410dc3e285c9a4317deb29745

PR #171, opened 2023-09-27
feat: add pre & post liquidation hooks

avniculae:
https://www.notion.so/voltz/Pre-Post-hooks-b08aeed29865472292c52096b83486f5?pvs=4


core/src/libraries/LiquidationBidPriorityQueue.sol
```diff
@@ -17,6 +17,7 @@ library LiquidationBidPriorityQueue {
 
     struct LiquidationBid {
         uint128 liquidatorAccountId;
+        address hookAddress;
```
0xZenus, line 20: Why is the hook address defined per bid?

core/src/modules/LiquidationModule.sol
```diff
@@ -70,6 +71,13 @@ contract LiquidationModule is ILiquidationModule {
         // grab the liquidatable account and check its existance
         Account.Data storage account = Account.exists(liquidatableAccountId);
 
+        if (liquidationBid.hookAddress != address(0)) {
```
avniculae, line 74: add todo

core/src/libraries/LiquidationBidPriorityQueue.sol
arturbeg: each liquidator may want to use custom hook implementation, e.g. one may want to do a flash loan, liquidate, withdraw and then return the flash loan, do you think there's an alternative way of approaching this?

core/src/interfaces/external/ILiquidationHook.sol
```diff
@@ -0,0 +1,41 @@
...
+You may obtain a copy of the License at
+
+https://github.com/Voltz-Protocol/v2-core/blob/main/core/LICENSE
+*/
+pragma solidity >=0.8.19;
+
+import {LiquidationBidPriorityQueue} from "../../libraries/LiquidationBidPriorityQueue.sol";
+import {IERC165} from "@voltz-protocol/util-contracts/src/interfaces/IERC165.sol";
+
+interface ILiquidationHook is IERC165 {
```
arturbeg, line 13: additionally the hook will need to be granted "ADMIN" access by the account to be able to withdraw and deposit on its behalf? worth adding it in natspec?

core/src/modules/LiquidationModule.sol
```diff
@@ -70,6 +81,17 @@ contract LiquidationModule is ILiquidationModule {
...
+            if (
+                ILiquidationHook(liquidationBid.hookAddress).preLiquidationHook(
+                    liquidatableAccountId,
+                    liquidationBid
+                ) != ILiquidationHook.preLiquidationHook.selector
+            ) {
+                revert InvalidPreLiquidationHookResponse();
```
arturbeg, line 91: what's this check trying to achieve, can it end up blocking the queue pop?

core/src/interfaces/external/ILiquidationHook.sol
avniculae: yes, that's correct.

core/src/modules/LiquidationModule.sol
avniculae: This is inside the `executeLiquidationBid` function, which is called using low-level `.call()` inside the `executeTopRankedLiquidationBid`. Therefore,  popping from the queue will not be blocked if this reverts.

avniculae: To answer the first question though, this is an extra check inspired from Uniswap v4 (see [this](https://github.com/Uniswap/v4-core/blob/0095e0848098c3e32e016eac6d2537b67aa47358/contracts/PoolManager.sol#L115))

core/src/interfaces/external/ILiquidationHook.sol
avniculae: [b4ea582](https://github.com/Voltz-Protocol/v2-core/pull/171/commits/b4ea582b8a03d7a9bece5ca2d59e5a8f508aeb82)

core/src/modules/LiquidationModule.sol
arturbeg: got it, thank you

arturbeg (approved):

PR #172, opened 2023-09-27
Feat/backstop-lps

avniculae:
https://www.notion.so/voltz/backstop-lps-premature-withdrawals-and-rewards-issue-ec535c070bfd4f2d818cafd7df20752c?pvs=4


core/src/libraries/actions/EditCollateral.sol
```diff
@@ -105,11 +143,60 @@ library EditCollateral {
...
+    /**
+     * @notice Backstop lp announces intention of withdrawal.
+     */
+    function announceBackstopLpWithdraw(uint128 accountId) internal {
+        Account.Data storage account = Account.exists(accountId);
+
+        CollateralPool.Data storage collateralPool = account.getCollateralPool();
+        uint128 backstopLpAccountId = collateralPool.backstopLPConfig.accountId;
+
+        if (backstopLpAccountId != accountId) {
```
0xZenus, line 173: why do we pass accountId if they only option is for it to be backstopLpAccountId?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -357,13 +357,21 @@ library AccountLiquidation {
             collateralPool.insuranceFundConfig.liquidationFee,
             liquidationPenalty
         );
-        uint256 backstopLPReward = mulUDxUint(
-            collateralPool.backstopLPConfig.liquidationFee,
-            liquidationPenalty
-        );
 
-        uint256 keeperReward = 0;
+        int256 backstopLpNetDepositsInUSD = backstopLpAccount.getMarginInfoByBubble(address(0)).netDeposits;
```
0xZenus, line 361: is it net deposits in here? Could you share the reference of how this algorithm should work?

core/src/libraries/actions/EditCollateral.sol
avniculae: thought about this as well; I am passing it so that the operation can be executed through the Execution Module (and so that I don't do the check in the Execution Module). Do you think I should do the check in the execution module?

core/src/libraries/account/AccountLiquidation.sol
avniculae: should be imDelta, check with @arturbeg

voltzprotocol (approved):

PR #175, opened 2023-09-28
feat: add dutch d parameter logic

avniculae:
https://www.notion.so/voltz/reward-logic-for-adl-order-propagation-needs-to-be-finalized-57fdd463dcf646548202410a4fe8aed2?pvs=4

### DIFF
--- a/core/src/libraries/account/AccountExposure.sol
+++ b/core/src/libraries/account/AccountExposure.sol
@@ -73,12 +73,12 @@ library AccountExposure {
             liquidationDelta: getExchangedQuantity(marginInfo.liquidationDelta, exchange.price, exchange.priceHaircut),
             dutchDelta: getExchangedQuantity(marginInfo.dutchDelta, exchange.price, exchange.priceHaircut),
             adlDelta: getExchangedQuantity(marginInfo.adlDelta, exchange.price, exchange.priceHaircut),
-            adlHealthInfo: Account.AdlHealthInformation({
+            dutchHealthInfo: Account.DutchHealthInformation({
                 rawMarginBalance: 
-                    getExchangedQuantity(marginInfo.adlHealthInfo.rawMarginBalance, exchange.priceHaircut, ZERO),
+                    getExchangedQuantity(marginInfo.dutchHealthInfo.rawMarginBalance, exchange.priceHaircut, ZERO),
                 rawLiquidationMarginRequirement:
                     getExchangedQuantity(
-                        marginInfo.adlHealthInfo.rawLiquidationMarginRequirement.toInt(), 
+                        marginInfo.dutchHealthInfo.rawLiquidationMarginRequirement.toInt(), 
                         exchange.priceHaircut, 
                         ZERO
                     ).toUint()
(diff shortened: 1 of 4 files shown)


core/src/libraries/account/AccountExposure.sol
```diff
@@ -71,7 +72,17 @@ library AccountExposure {
             maintenanceDelta: getExchangedQuantity(marginInfo.maintenanceDelta, exchange.price, exchange.priceHaircut),
             liquidationDelta: getExchangedQuantity(marginInfo.liquidationDelta, exchange.price, exchange.priceHaircut),
             dutchDelta: getExchangedQuantity(marginInfo.dutchDelta, exchange.price, exchange.priceHaircut),
-            adlDelta: getExchangedQuantity(marginInfo.adlDelta, exchange.price, exchange.priceHaircut)
+            adlDelta: getExchangedQuantity(marginInfo.adlDelta, exchange.price, exchange.priceHaircut),
+            adlHealthInfo: Account.AdlHealthInformation({
```
arturbeg, line 76: isn't it dutch health info?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -393,9 +394,30 @@ library AccountLiquidation {
         }
     }
 
+    function computeDutchHealth(Account.Data storage self) internal view returns (UD60x18) {
+        Account.MarginInfo memory marginInfo = self.getMarginInfoByBubble(address(0));
+        // adl health info values are in USD (and therefore represented with 18 decimals)
+        UD60x18 health = ud(marginInfo.adlHealthInfo.rawMarginBalance.toUint()).div(
```
arturbeg, line 400: since we're converting raw margin balance to uint, is it necessary to have a check for raw margin balance >= 0?

arturbeg: is it worth using divUintUD from prbmath helper (tho it only works with uint and ud pair)

arturbeg (approved):

core/src/libraries/account/AccountLiquidation.sol
avniculae: Regarding first point: if raw margin balance is < 0, then `toUint()` will revert. Do you have another behaviour in mind?
Regarding second point: yeah, the prbmath helper is just for int/uint - ud/sd pairs.

avniculae: Regarding first point: if raw margin balance is < 0, then toUint() will revert. Do you have another behaviour in mind? Regarding second point: yeah, the prbmath helper is just for int/uint - ud/sd pairs.

avniculae: Shall I change anything here?

core/src/libraries/account/AccountExposure.sol
avniculae: Addressed: [0928393](https://github.com/Voltz-Protocol/v2-core/pull/175/commits/0928393ae15afd96cad0ad36a08c4c68de08d6de)

core/src/libraries/account/AccountLiquidation.sol
arturbeg: hmm, i think it's okay since if margin balance < 0 then it must be below the adl threshold at that point?

PR #178, opened 2023-09-29
Feat/integration test c

0xZenus:


exchanges/vamm/src/libraries/ticks/Tick.sol
```diff
@@ -260,6 +252,14 @@ library Tick {
                 info.trackerQuoteTokenGrowthOutsideX128
             );
 
+        info.trackerQuoteTokenGrowthOutsideX128 =
```
ioanaroceanu2000, line 255: can you explain this change?

0xZenus: Yes. The issue was that when marking to market, the last liquidity index and timestamps were not initialised (were 0, 0) and if we updated base and quote before, and then marking to market, there was some accrued interest `base * (current_liquidity_index - 0) + quote * (current_timestamp - 0)` which was wrong. Hence, we needed to mark-to-market before because at that time, base and quote were zeroes and even though we accumulated `base * (current_liquidity_index - 0) + quote * (current_timestamp - 0)`, that was 0, because base and quote are 0 initially.

ioanaroceanu2000: got it thanks!

ioanaroceanu2000 (approved):

PR #179, opened 2023-10-02
Integration test D

0xZenus:


exchanges/vamm/src/libraries/vamm-utils/VammConfiguration.sol
```diff
@@ -97,7 +98,7 @@ library VammConfiguration {
         // need to make sure the currently-held invariant that "current tick is always within the allowed tick range"
         // does not have unwanted consequences
         if(
-            minTickAllowed < TickMath.MIN_TICK_LIMIT || maxTickAllowed > TickMath.MAX_TICK_LIMIT ||
+            minTickAllowed < VammTicks.DEFAULT_MIN_TICK || maxTickAllowed > VammTicks.DEFAULT_MAX_TICK ||
```
ioanaroceanu2000, line 101: why not the min tick allowed? it gives more flexibility than the default

exchanges/vamm/src/libraries/vamm-utils/VammTicks.sol
```diff
@@ -75,18 +80,25 @@ library VammTicks {
...
+            getPriceFromTick(DEFAULT_MIN_TICK).gt(maxPrice) 
+                ? getTickFromPrice(maxPrice) 
+                : DEFAULT_MIN_TICK;
         
-        dynamicMinTick = getTickFromPrice(maxPrice);
-        dynamicMaxTick = getTickFromPrice(minPrice);
+        dynamicMaxTick = 
+            getPriceFromTick(DEFAULT_MAX_TICK).lt(minPrice)
+                ? getTickFromPrice(minPrice)
+                : DEFAULT_MAX_TICK;
```
ioanaroceanu2000, line 101: I would apply the same strategy for twap instead of returning 0 when spreads and slippage are too big, wdyt?

integration/test/scenarios/ScenarioB.sol
```diff
@@ -214,16 +214,6 @@ contract ScenarioA is ScenarioSetup, AssertionHelpers, Actions, Checks {
                 assertAlmostEq(executedQuote, int256(45326575), 1e6, "executedQuote1"); 
                 assertEq(annualizedNotional, -505000000, "annualizedNotional1");
             }
-            // twap checks
-            {
```
ioanaroceanu2000, line 218: why removed?

integration/test/scenarios/ScenarioD.sol
```diff
@@ -0,0 +1,553 @@
...
+        }
+
+        {
+            uint256 twap = getAdjustedTwap(marketId, maturityTimestamp, 0);
+            assertEq(twap, 47446712689052953, "twap after 10 days when order = 0");
+        }
+
+        {
+            uint256 twap = getAdjustedTwap(marketId, maturityTimestamp, 100e18);
+            assertEq(twap, 47446712689052953, "twap after 10 days when order = 100");
```
ioanaroceanu2000, line 214: is it expected to have the same value as the above even if the order size is 100?

integration/test/scenarios/ScenarioB.sol
0xZenus: we added proper test for twap (test D).

exchanges/vamm/src/libraries/vamm-utils/Twap.sol
```diff
@@ -39,37 +37,66 @@ library Twap {
...
-                return ZERO;
-            }
-            if (priceImpactAsFraction.gte(UNIT)) {
-                // The model suggests that the price will drop below zero after price impact
-                return ZERO;
-            }
-            geometricMeanPrice = geometricMeanPrice.mul(UNIT.sub(priceImpactAsFraction)).sub(spreadImpactDelta);
+        if (priceImpactAsFraction.gte(UNIT_ud)) {
+            // The model suggests that the price will drop below zero after price impact
+            return ZERO_ud;
```
ioanaroceanu2000, line 76: twap

exchanges/vamm/src/libraries/vamm-utils/Twap.sol
```diff
@@ -39,37 +37,66 @@ library Twap {
...
+        if (orderSize.eq(ZERO_sd)) {
+            return price;
+        }
+
+        if (orderSize.gt(ZERO_sd)) {
+            return price.add(spread);
+        }
+
+        if (price.lte(spread)) {
+            return ZERO_ud;
```
ioanaroceanu2000, line 96: return MAX tick instead

exchanges/vamm/src/libraries/vamm-utils/VammConfiguration.sol
0xZenus: The old min_tick was the tick used by Uniswap (which was so huge in absolute value that the corresponding sqrt price was zero).

exchanges/vamm/src/libraries/vamm-utils/Twap.sol
0xZenus: this function returns the price, it doesn't deal with ticks so it'd be inconsistent to return ticks here.

exchanges/vamm/src/libraries/vamm-utils/VammTicks.sol
0xZenus: answered to your comments.

ioanaroceanu2000 (approved):

PR #182, opened 2023-10-03
Feat/adl

avniculae:
- https://www.notion.so/voltz/adl-timer-implementation-6dc29f1085ad4d7091271fe090a1541b?pvs=4
- https://www.notion.so/voltz/adjustments-to-the-backstop-liquidation-function-2ae175587f8f4087916c189e082b672b?pvs=4

### DIFF
--- a/core/src/libraries/account/AccountLiquidation.sol
+++ b/core/src/libraries/account/AccountLiquidation.sol
@@ -148,13 +148,13 @@ library AccountLiquidation {
      * @dev Checks if an account is insolvent in a given bubble (assuming auto-exchange is exhausted!!!)
      * and returns the shortfall
      */
-    function isInsolvent(Account.Data storage self, address collateralType) private view returns (bool, int256) {
+    function isInsolvent(Account.Data storage self, address collateralType) private view returns (bool) {
         // todo: note, doing too many redundunt calculations, can be optimized
         // todo: consider reverting if address(0) is provided as collateralType
         // consider baking this function into the backstop lp function if it's not used anywhere else
 
         Account.MarginInfo memory marginInfo = self.getMarginInfoByBubble(collateralType);
-        return (marginInfo.collateralInfo.marginBalance < 0, marginInfo.collateralInfo.marginBalance);
+        return marginInfo.collateralInfo.marginBalance < 0;
     }
 
 
@@ -557,7 +557,6 @@ library AccountLiquidation {
 
         // revert if account is not below adl margin requirement
         Account.MarginInfo memory marginInfo = self.getMarginInfoByBubble(address(0));
-
         if (marginInfo.adlDelta > 0) {
             revert AccountNotBelowADL(self.id, marginInfo);
         }
@@ -565,14 +564,11 @@ library AccountLiquidation {
         // todo: layer in backstop lp & keeper rewards
         // todo: make sure backstop lp capacity is exhausted before proceeding to adl
 
-        (bool _isInsolvent, int256 marginBalance) = isInsolvent(self, quoteToken);
-
         CollateralPool.Data storage collateralPool = self.getCollateralPool();
-
         uint256 shortfall = 0;
 
+        bool _isInsolvent = isInsolvent(self, quoteToken);
         if (!_isInsolvent) {
-
             Account.Data storage backstopLpAccount = Account.exists(collateralPool.backstopLPConfig.accountId);
 
             // execute backstop lp liquidation orders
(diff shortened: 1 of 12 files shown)


core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -148,13 +148,13 @@ library AccountLiquidation {
...
      */
-    function isInsolvent(Account.Data storage self, address collateralType) private view returns (bool, int256) {
+    function isInsolvent(Account.Data storage self, address collateralType) private view returns (bool) {
         // todo: note, doing too many redundunt calculations, can be optimized
         // todo: consider reverting if address(0) is provided as collateralType
         // consider baking this function into the backstop lp function if it's not used anywhere else
 
         Account.MarginInfo memory marginInfo = self.getMarginInfoByBubble(collateralType);
-        return (marginInfo.collateralInfo.marginBalance < 0, marginInfo.collateralInfo.marginBalance);
+        return marginInfo.collateralInfo.marginBalance < 0;
```
arturbeg, line 157: think when assessing for insolvency in a given bubble we no longer need to apply haircuts?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -549,73 +549,165 @@ library AccountLiquidation {
 
     function executeBackstopLiquidation(
         Account.Data storage self,
-        uint128 liquidatorAccountId,
+        uint128 liquidatorAccountId, // todo: why do we pass this here?
```
arturbeg, line 552: this is the keeper that triggers the function, will be necessary once we propagate keeper rewards

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -549,73 +549,165 @@ library AccountLiquidation {
...
-                );
+        uint256[] memory markets = self.activeMarketsPerQuoteToken[quoteToken].values();
+        for (uint256 i = 0; i < markets.length && !leftExposure; i++) {
+            uint128 marketId = markets[i].to128();
+            Market.Data storage market = Market.exists(marketId);
+
+            Account.MakerMarketExposure[] memory exposures = 
+                market.getAccountTakerAndMakerExposures(self.id);
+
+            if (exposures.length > 0) {
```
arturbeg, line 605: is this check sufficient, it might be the case that the exposure object is still returned since there's non-zero pnl components within it unless i'm missing details of getAccountTakerAndMakerExposures implementation

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -549,73 +549,165 @@ library AccountLiquidation {
...
-        } else {
-            Account.Data storage insuranceFundAccount = Account.exists(collateralPool.insuranceFundConfig.accountId);
-            int256 insuranceFundCoverAvailable = insuranceFundAccount.getAccountNetCollateralDeposits(quoteToken)
-            - collateralPool.insuranceFundUnderwritings[quoteToken].toInt();
-
-            uint256 insuranceFundDebit = (-marginBalance).toUint();
-            if (insuranceFundCoverAvailable + marginBalance < 0) {
-                shortfall = (marginBalance-insuranceFundCoverAvailable).toUint();
-                insuranceFundDebit = insuranceFundCoverAvailable > 0 ? (-insuranceFundCoverAvailable).toUint() : 0;
+            backstopLpAccount.imBufferCheck(address(0));
```
arturbeg, line 612: might be worth leaving a todo since i think we're double calculating margin requirements even though we just need to in theory apply a single extra multiplier to the lm to get im buffer check :?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -549,73 +549,165 @@ library AccountLiquidation {
...
+
+            for (uint256 i = 0; i < markets.length; i++) {
+                uint128 marketId = markets[i].to128();
+                Market.exists(marketId).executeADLOrder({
+                    liquidatableAccountId: self.id,
+                    inLoss: true,
```
arturbeg, line 618: why do we pass inLoss in here?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -549,73 +549,165 @@ library AccountLiquidation {
...
+
+        if (insuranceFundCoverAvailable + marginInfo.collateralInfo.marginBalance > 0) {
+            collateralPool.updateInsuranceFundUnderwritings(quoteToken, (-marginInfo.collateralInfo.marginBalance).toUint());
+            
+            // adl maturities with negative upnl at market price
+            for (uint256 i = 0; i < markets.length; i++) {
+                uint128 marketId = markets[i].to128();
+                Market.exists(marketId).executeADLOrder({
+                    liquidatableAccountId: self.id,
+                    inLoss: true,
```
arturbeg, line 665: at this stage everything that's left is by definition in loss ?

core/src/storage/Account.sol
```diff
@@ -430,4 +440,15 @@ library Account {
         }
     }
 
+    function imBufferCheck(Data storage self, address collateralType)
```
arturbeg, line 443: does this need to be a separate function given it's only used in one place?

exchanges/vamm/src/storage/LPPosition.sol
```diff
@@ -74,7 +74,7 @@ library LPPosition {
     }
 
     /**
-     * @dev Loads the LPPosition object for the given position Id
+     * @dev Loads the LPPosigettion object for the given position Id
```
arturbeg, line 77: typo?

avniculae: Thanks, addressed

core/src/storage/Account.sol
avniculae: Refactored to `imAndImBufferCheck` to not duplicate the margin info computation

core/src/libraries/account/AccountLiquidation.sol
avniculae: yes, that's correct

core/src/libraries/account/AccountLiquidation.sol
avniculae: Addressed

core/src/libraries/account/AccountLiquidation.sol
avniculae: Correct, addressed

core/src/libraries/account/AccountLiquidation.sol
avniculae: Makes sense, thanks

core/src/libraries/account/AccountLiquidation.sol
avniculae: Correct, addressed.

arturbeg (approved):

PR #191, opened 2023-10-12
VAMM Clean-up Part 2

0xZenus:


exchanges/vamm/src/libraries/vamm-utils/VammHelpers.sol
```diff
@@ -129,26 +128,12 @@ library VammHelpers {
...
-        }
-
-        int256 exposure = mulUDxInt(exposureFactor, baseTokenDelta);
-
-        quoteTokenDelta = mulUDxInt(averagePriceWithSpread, -exposure);
+    function calculatePrice(uint256 base, uint256 unbalancedQuote) internal pure returns (UD60x18) {
+        return ud(unbalancedQuote).div(ud(base)).div(convert(100));
+    }
+
+    function applySpread(UD60x18 price, UD60x18 spread, bool isLPLong) internal pure returns (UD60x18) {
```
ioanaroceanu2000: don't we already have this function in Twap.sol?

0xZenus: That's true. That function requires orderSizeWad (and I don't have tokenDecimals here to turn unfilledBase). Let's keep it as it is for now and we'll think more about it.

PR #195, opened 2023-10-13
Bankruptcy price & scoring function

avniculae:


products/dated-irs/src/storage/Market.sol
```diff
@@ -182,6 +192,14 @@ library Market {
          */
         mapping(uint32 maturityTimestamp => uint256 notional) notionalTracker;
 
+        /**
+         * The phi value used to compute percentual slippage
+         */
+        mapping(uint32 maturityTimestamp => UD60x18 phi) phis;
```
arturbeg, line 198: can phi and beta be a struct?

products/dated-irs/src/modules/MarketManagerIRSModule.sol
```diff
@@ -169,6 +175,35 @@ contract MarketManagerIRSModule is IMarketManagerIRSModule {
...
+        ) = abi.decode(inputs, (uint32, int256));
+
+        int256 annualizedExposure = ExposureHelpers.baseToAnnualizedExposure(
+            baseToBeLiquidated,
+            marketId,
+            maturityTimestamp
+        );
+
+        Market.Data storage market = Market.exists(marketId);
+        annualizedExposureWad = DecimalMath.changeDecimals(
```
arturbeg, line 194: can you remind me why we use decimal math here instead of prb math ?

exchanges/vamm/src/libraries/vamm-utils/VammConfiguration.sol
```diff
@@ -83,18 +81,11 @@ library VammConfiguration {
...
-        self.mutableConfig.priceImpactPhi = config.priceImpactPhi;
-        self.mutableConfig.spread = config.spread;
-
-        setMinAndMaxTicks(self, config.minTickAllowed, config.maxTickAllowed);
+        self.mutableConfig = config;
+        propagateMinAndMaxTicks(self, config.minTickAllowed, config.maxTickAllowed);
     }
 
-    function setMinAndMaxTicks(
+    function propagateMinAndMaxTicks(
```
arturbeg, line 88: why's it propagate instead of set now?

products/dated-irs/src/storage/Market.sol
avniculae: They can. Personally I think we should create a struct for all configs per market & maturity, including risk params. wdyt?

exchanges/vamm/src/libraries/vamm-utils/VammConfiguration.sol
avniculae: The set operation is done above, where we store the entire struct to the storage. However, after we store in in storage, we need to call this function to double-check numbers and to propagate the values to global trackers as well

products/dated-irs/src/storage/Market.sol
arturbeg: hmm, yeah that makes sense, what other configs are there, can you remind me?

products/dated-irs/src/libraries/ExposureHelpers.sol
```diff
@@ -40,6 +40,18 @@ library ExposureHelpers {
 
     uint256 internal constant SECONDS_IN_DAY = 86400;
 
+    function getPercentualSlippage(uint128 marketId, uint32 maturityTimestamp, int256 annualizedExposureWad) internal view returns (UD60x18) {
```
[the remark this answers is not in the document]
avniculae: addressed

arturbeg (approved):

PR #199, opened 2023-10-18
Feat/inspecting dated irs

0xZenus:


core/src/storage/Market.sol
```diff
@@ -194,10 +194,12 @@ library Market {
     function validateLiquidationOrder(
         Data storage self,
         uint128 liquidatableAccountId,
+        uint128 liquidatorAccountId,
         bytes memory inputs
     ) internal view {
         IMarketManager(self.marketManagerAddress).validateLiquidationOrder(
             liquidatableAccountId,
+            liquidatorAccountId,
```
ioanaroceanu2000, line 202: why do we need liquidatorAccountId to validate?

products/dated-irs/src/modules/MarketManagerIRSModule.sol
```diff
@@ -186,17 +180,26 @@ contract MarketManagerIRSModule is IMarketManagerIRSModule {
...
     {
-        (uint32 maturityTimestamp, int256 baseAmountToBeLiquidated) = abi.decode(inputs, (uint32, int256));
+        (uint32 maturityTimestamp, int256 baseAmountToBeLiquidated, uint160 priceLimit) =
+            abi.decode(inputs, (uint32, int256, uint160));
 
         ExecuteLiquidationOrder.validateLiquidationOrder(
-            liquidatableAccountId, marketId, maturityTimestamp, baseAmountToBeLiquidated
+            LiquidationOrderParams({
+                liquidatableAccountId: liquidatableAccountId,
+                liquidatorAccountId: liquidatorAccountId,
```
ioanaroceanu2000, line 197: can we replace by 0?

core/src/storage/Market.sol
0xZenus: we don't but I want to keep consistency between validation and execution, such that both receive the same LiquidationParams structure.

PR #200, opened 2023-10-18
Feat/liq rewards

avniculae:
- https://www.notion.so/voltz/reward-logic-for-adl-execution-cbed5f7f312440b18af564bfaeaa98ff?pvs=4
- https://www.notion.so/voltz/reward-logic-for-adl-order-propagation-needs-to-be-finalized-57fdd463dcf646548202410a4fe8aed2?pvs=4
- https://www.notion.so/voltz/non-refundable-flat-deposit-when-a-liquidation-bid-is-submitted-to-discuss-efcc134a8fd34166aaa1b8cf5d5044f1?pvs=4


core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -799,5 +807,46 @@ library AccountLiquidation {
...
         }
     }
+
+    function propagateADLOrder(
+        Account.Data storage self,
+        uint128 marketId, 
+        uint128 keeperAccountId, 
+        uint32 maturityTimestamp, 
+        bool isLong
+    ) internal {
```
avniculae, line 817: shall we check that keeper account id and self.id share the same collateral pool? @0xZenus

core/src/interfaces/external/IMarketManager.sol
```diff
@@ -128,6 +128,14 @@ interface IMarketManager is IERC165 {
         int256 realBalanceAndIF
     ) external;
 
+    // todo: add natspec
+    function propagateADLOrder(
+        uint128 accountId, 
+        uint128 marketId, 
+        uint32 maturityTimestamp,
```
arturbeg, line 135: hmm, since not all instruments may have markets / might need to have different inputs for adl order propagation, isn't it cleaner to have bytes inputs instead of the maturity timestamp, maybe even isLong similar to how we handle other IMarketManager functions and then decode them in the instrument to provide more flexibility?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -613,18 +679,45 @@ library AccountLiquidation {
                     realBalanceAndIF: 0
                 });
             }
+
+            vars.rawLMRAfterAdl = self.getMarginInfoByBubble(
```
arturbeg, line 683: this should be zero right? maybe can save some gas by avoiding this calculation unless there's a risk of missing smth?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -638,13 +731,11 @@ library AccountLiquidation {
...
-        Account.Data storage insuranceFundAccount = 
-            Account.exists(collateralPool.insuranceFundConfig.accountId);
-        int256 insuranceFundCoverAvailable = 
-            insuranceFundAccount.getAccountNetCollateralDeposits(quoteToken) - 
-                collateralPool.insuranceFundUnderwritings[quoteToken].toInt();
-
-        if (insuranceFundCoverAvailable + marginInfo.collateralInfo.marginBalance > 0) {
+        uint256 insuranceFundCoverAvailable = collateralPool.getAvailableInsuranceFundCover({
+            quoteToken: quoteToken, 
+            preserveMinThreshold: true
```
arturbeg, line 736: is there ever an instance when we don't want to preserve the min threshold, i think during propagation?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -638,13 +731,11 @@ library AccountLiquidation {
...
+        });
+        if (insuranceFundCoverAvailable.toInt() + marginInfo.collateralInfo.marginBalance > 0) {
```
arturbeg, line 738: hmm, so the cover available is in quote terms, but the margin balance is in dollar terms, shouldn't we get the margin info for the quote token? assuming auto-exchange has taken place

core/src/libraries/account/AccountLiquidation.sol
arturbeg: yes, for consistency, i think there's a function in the account that does this

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -718,5 +807,46 @@ library AccountLiquidation {
...
+    ) internal {
+        CollateralPool.Data storage collateralPool = self.getCollateralPool();
+        Market.Data storage market = Market.exists(marketId);
+
+        uint256 insuranceFundCoverAvailable = collateralPool.getAvailableInsuranceFundCover({
+            quoteToken: market.quoteToken, 
+            preserveMinThreshold: false
+        });
+
+        if (insuranceFundCoverAvailable < collateralPool.riskConfig.liquidationConfiguration.adlPropagationKeeperFee) {
```
arturbeg, line 826: hmm, i think since the propagation keeper fee is flat, it should be unique for each collateral asset, so in my head it makes more sense to have this parameter as part of collateral configuration rather than collateral pool risk config?

core/src/storage/CollateralPool.sol
```diff
@@ -192,6 +209,11 @@ library CollateralPool {
          * @dev Pool's insurance fund account ID
          */
         uint128 accountId;
+        /**
+         * @dev Minimum funds threshold that must be left in insurance fund
+         * for covering keeper rewards for adl propagation.
+         */
+        uint256 minInsuranceFundThreshold;
```
arturbeg, line 216: hmm, why's the min insturance fund threshold in dutch configuration?

core/src/storage/CollateralPool.sol
```diff
@@ -289,6 +311,25 @@ library CollateralPool {
...
+        Data storage self, 
+        address quoteToken,
+        bool preserveMinThreshold
+    ) internal view returns (uint256) {
+        Account.Data storage insuranceFundAccount = Account.exists(self.insuranceFundConfig.accountId);
+        int256 insuranceFundBalance = insuranceFundAccount.getAccountNetCollateralDeposits(quoteToken);
+
+        insuranceFundBalance -= self.insuranceFundUnderwritings[quoteToken].toInt();
+        if (preserveMinThreshold) {
+            insuranceFundBalance -= self.insuranceFundConfig.minInsuranceFundThreshold.toInt();
```
arturbeg, line 324: also, i think the min insurance fund cover should be unique for each quote token, as part of collateral configuration?

core/src/libraries/account/AccountLiquidation.sol
avniculae: yes, during propagation

core/src/libraries/account/AccountLiquidation.sol
arturbeg: here's an example https://github.com/Voltz-Protocol/v2-core/blob/cd3751f515c51f51947c9e00328a6262c28837c1/core/src/storage/CollateralConfiguration.sol#L96, auto-exchange thresholds a unique for each collateral asset to avoid the need to care about dollar conversions, decimals, and provides further flexibility

products/dated-irs/src/modules/MarketManagerIRSModule.sol
```diff
@@ -339,7 +339,7 @@ contract MarketManagerIRSModule is IMarketManagerIRSModule {
     }
 
     /**
-     * @inheritdoc IMarketManagerIRSModule
+     * @inheritdoc IMarketManager
      */
     function propagateADLOrder(
```
avniculae, line 344: should block calls if msg sender is note core

arturbeg: yeah, good point

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -241,8 +253,22 @@ library AccountLiquidation {
...
         validateLiquidationBid(self, liquidatorAccount, liquidationBid);
-        uint256 liquidationBidRank = computeLiquidationBidRank(self, liquidationBid);
+
         CollateralPool.Data storage collateralPool = self.getCollateralPool();
+        CollateralConfiguration.Data storage collateralConfig = CollateralConfiguration.exists(
+            collateralPool.id, 
+            liquidationBid.quoteToken
+        );
+
+        EditCollateral.deposit({
```
arturbeg, line 263: hmm, why should we deposit here? should't we just take collateralConfig.baseConfig.bidSubmissionFee from the net deposits of the account and then as you do below credit it to the insurance fund?

core/src/modules/liquidation/CommonLiquidationModule.sol
```diff
@@ -0,0 +1,126 @@
...
+        account.distributeLiquidationPenalty(
+            liquidatorAccount,
+            liquidationPenalty,
+            liquidationBid.quoteToken,
+            bidSubmissionKeeperId
+        );
+
+        liquidatorAccount.imCheck();
+
+        // todo: should in theory revert if the account is insolvent (& insurance fund
```
arturbeg, line 110: i think this todo is now redudunt because we have is above adl check right?

avniculae: that's correct, we have the `betweenAdlAndLmCheck` check

avniculae: Addressed

core/src/libraries/account/AccountLiquidation.sol
avniculae: Addressed

products/dated-irs/src/modules/MarketManagerIRSModule.sol
avniculae: Addressed

core/src/storage/CollateralPool.sol
avniculae: minInsuranceFundThreshold is in InsuranceFundConfig, can you check again? (github might hide parts of the code in the diff)

core/src/libraries/account/AccountLiquidation.sol
avniculae: Addressed

core/src/interfaces/external/IMarketManager.sol
avniculae: You are right, addressed

core/src/libraries/account/AccountLiquidation.sol
avniculae: hmm is this the case for the solvent case as well? I thought LMR might be above zero in this case

core/src/libraries/account/AccountLiquidation.sol
avniculae: Addressed

core/src/storage/CollateralPool.sol
avniculae: In any case, I moved it to collateral configuration as you suggested

core/src/libraries/account/AccountLiquidation.sol
avniculae: Addressed

core/src/storage/CollateralPool.sol
avniculae: Addressed

core/src/storage/CollateralPool.sol
arturbeg: nice

arturbeg (approved):

PR #205, opened 2023-10-19
Splitting taker maker exposures

0xZenus:


core/src/interfaces/IMarketManagerModule.sol
```diff
@@ -37,15 +37,6 @@ interface IMarketManagerModule {
     /// @notice returns the id of the last created market
     function getLastCreatedMarketId() external returns (uint128);
 
-
-    /// @notice returns account taker and maker exposures for a given market and account
```
0xZenus, line 41: I removed this function from core since it's already in the dated irs in the same form.

products/dated-irs/src/storage/Portfolio.sol
```diff
@@ -195,74 +195,103 @@ library Portfolio {
...
+            Market.MarketMaturityConfiguration memory marketMaturityConfig =
+                market.marketMaturityConfigs[maturityTimestamp];
+
+            if (cachedUnfilledBalances[i - 1].baseLong == 0 && cachedUnfilledBalances[i - 1].baseShort == 0) {
+                continue;
+            }
+
+            unfilledExposures[size].exposureComponents = Account.UnfilledExposureComponents({
+                long: ExposureHelpers.getExposureComponents(
+                    cachedUnfilledBalances[i - 1].baseLong.toInt(),
```
0xZenus, line 241: do we use here the same sign for both long and short exposures? @arturbeg

products/dated-irs/src/libraries/ExposureHelpers.sol
```diff
@@ -113,13 +113,13 @@ library ExposureHelpers {
     {
         uint32 timestamp = Time.blockTimestampTruncated();
 
-        if (maturityTimestamp <= timestamp) {
+        if ((base == 0 && quote == 0) || maturityTimestamp <= timestamp) {
```
arturbeg, line 116: hmm, i'm not sure quote == 0 is necessary, base == 0 is sufficient to return 0?

products/dated-irs/src/libraries/ExposureHelpers.sol
```diff
@@ -176,108 +163,45 @@ library ExposureHelpers {
...
-        unfilledExposureComponents = new Account.UnfilledExposureComponents[](2);
-        uint256 timeToMaturityInSeconds = uint256(maturityTimestamp) - block.timestamp;
-
-        if (unfilledBaseLong != 0) {
-            uint256 unfilledNotionalLong = mulUDxUint(exposureFactor, unfilledBaseLong);
+        exposureComponents = new int256[](2);
 
-            (int256 unfilledExposureShortRate, int256 unfilledExposureSwapRate) =
-                decoupleExposures(unfilledNotionalLong.toInt(), tenorInSeconds, timeToMaturityInSeconds);
+        // todo: understand how tenorInSeconds compares to SECONDS_IN_DAY
```
arturbeg, line 178: tenorInSeconds is the original duration of the maturity at the time of maturity inception if that makes sense?

products/dated-irs/src/libraries/ExposureHelpers.sol
0xZenus: This is uPnL. Think of the case where you've done two swaps (one with base x and one with base -x (unwind)). You're left with some quote that will contribute to the uPnL because locked PnL is part of uPnL for us at this moment.

products/dated-irs/src/libraries/ExposureHelpers.sol
0xZenus: thanks for the explanation.

PR #206, opened 2023-10-19
Partial data types and exposure factor fetched once

0xZenus:


core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -607,7 +608,7 @@ library AccountLiquidation {
         if (!_isInsolvent) {
             executeSolventBackstopLiquidation(self, keeperAccountId, quoteToken, backstopLPLiquidationOrders);
         } else {
-            executeInsolventADLLiquidation(self, keeperAccountId, quoteToken);
+            executeInsolventADLLiquidation(self, keeperAccountId, quoteToken, marginInfo);
```
avniculae, line 611: margin info shouldn't be passed here since we reverted buggy optimisation in the insolvent flow. can you revert this please (my branch fixed this bug)

avniculae (approved):

PR #208, opened 2023-10-20
Update ExchangeStore.sol: Usage of 'immutable' for keccak variables

Aryan9592:
It's always advisable to use 'immutable' for keccak variables rather than 'constant'
Refer to this for a better explanation -> https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#change-constant-to-immutable-for-keccak-variables


Aryan9592:
It be better if you could told me the reason why this PR is not suitable. At least that would be a good approach before closing a PR. Ain't it?

PR #211, opened 2023-10-23
feat: c percentage

ioanaroceanu2000:


core/src/libraries/PriceHelpers.sol
```diff
@@ -0,0 +1,23 @@
+/*
+Licensed under the Voltz v2 License (the "License"); you 
+may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://github.com/Voltz-Protocol/v2-core/blob/main/core/LICENSE
+*/
+pragma solidity >=0.8.19;
```
arturbeg, line 8: shall we just push this into the collateral configuration library?

core/src/libraries/account/AccountAutoExchange.sol
```diff
@@ -115,22 +117,32 @@ library AccountAutoExchange {
...
                 quoteType,
                 collateralPool.riskConfig.riskMultipliers
             );
 
         {
-
             // mismatched margin coverage check
-
             if ( (overallMarginInfo.maintenanceDelta < 0) && (marginInfo.liquidationDelta < 0) ) {
+                UD60x18 bubbleLossPercentageToCover = calculatePercentageOfCollateralToBeCovered(
```
arturbeg, line 131: if the account overall is above lm, we should be able to assume c is 1 for gas reasons  i think https://discord.com/channels/896685581487210577/1005077803563229285/1166048174180401225

core/src/libraries/account/AccountAutoExchange.sol
```diff
@@ -332,4 +344,60 @@ library AccountAutoExchange {
...
+                tokens[i],
+                riskMultipliers
+            );
+
+            CollateralConfiguration.ExchangeInfo memory exchangeInfo =
+                CollateralConfiguration.getExchangeInfo(collateralPoolId, tokens[i], address(0));
+
+            int256 shortfall;
+            int256 surplus;
+            if (marginInfo.liquidationDelta < marginInfo.collateralInfo.realBalance) {
```
arturbeg, line 372: worth cleaning up to see if we can avoid nested if statements and not have to update the shortfall if it's zero / not have to update surplus if it's zero

core/src/libraries/account/AccountAutoExchange.sol
ioanaroceanu2000: if the bubble is below LMR (which is the case here after this condition `marginInfo.liquidationDelta < 0`) then we should calculate the % to be covered of this loss.
if the bubble is above LMR, we don't enter this branch, no need to calculate any c

core/src/libraries/account/AccountAutoExchange.sol
ioanaroceanu2000: yes, not very nice indeed. refactored in 06f9a8e566556822455fe9d3cb19b57d50574be1

core/src/libraries/PriceHelpers.sol
ioanaroceanu2000: we call this function (`getExchangedQuantity`) 24 times, I'd rather have it in a utils library than in a storage one

arturbeg: sure

core/src/libraries/account/AccountAutoExchange.sol
arturbeg: hmm, i was actually thinking overMarginInfo.liquidationDelta < 0 as a further optimization, but worth checking and auditing with AA as well for completeness

arturbeg (approved):

PR #213, opened 2023-10-24
Feat/core int test

0xZenus:


core/src/libraries/actions/EditCollateral.sol
```diff
@@ -143,14 +143,17 @@ library EditCollateral {
...
-        
-        /// Check if account if backstop lp. If it is, make sure that 
-        /// withdraw period is active.
-        CollateralPool.Data storage collateralPool = account.getCollateralPool();
-        uint128 backstopLpAccountId = collateralPool.backstopLPConfig.accountId;
-        if (backstopLpAccountId == accountId) {
-            if (!Timer.loadOrCreate(backstopLpTimerId(backstopLpAccountId)).isActive()) {
-                revert BackstopLpWithdrawPeriodInactive(backstopLpAccountId, block.timestamp);
+
+        if (account.hasCollateralPool()) {
```
ioanaroceanu2000, line 147: what does this mean? and if the account doesn't have a collateral pools, does that mean it has no collateral?

core/src/modules/AccountModule.sol
```diff
@@ -10,7 +10,7 @@ pragma solidity >=0.8.19;
 import { AccountPermissions, MarginInfo } from "../libraries/DataTypes.sol";
 import {IAccountModule} from "../interfaces/IAccountModule.sol";
 import {Account} from "../storage/Account.sol";
-import {AccessPassConfiguration} from "../storage/AccessPassConfiguration.sol";
+import {AccountExposure} from "../libraries/account/AccountExposure.sol";
```
ioanaroceanu2000, line 13: why added?

core/src/modules/ExecutionModule.sol
```diff
@@ -81,10 +90,8 @@ contract ExecutionModule is IExecutionModule {
     ) internal returns (bytes memory output) {
 
         Account.Data storage account = Account.exists(accountId);
-        if (account.getCollateralPool().id != Market.exists(command.marketId).getCollateralPool().id) {
```
ioanaroceanu2000, line 84: why did we skip this?

core/src/modules/liquidation/CommonLiquidationModule.sol
```diff
@@ -25,6 +26,7 @@ import { mulUDxUint } from "@voltz-protocol/util-contracts/src/helpers/PrbMathHe
 
 contract CommonLiquidationModule is ICommonLiquidationModule {
     using Account for Account.Data;
+    using AccountExposure for Account.Data;
```
ioanaroceanu2000, line 29: why added?

integration/test/scenarios/core_scenarios/ScenarioA.sol
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        // maker order
+        {
+            make({
+                accountId: 1,
+                user: user1,
+                marketId: 1,
+                exchangeId: 1,
+                inputs: abi.encode(maturityTimestamp, -19_500, -11_040, 10e18)
```
ioanaroceanu2000, line 202: might be helpful to signal what these inputs are (either by comments or by encoding the input within the helper function)

core/src/libraries/actions/EditCollateral.sol
0xZenus: This check here is for backstop LP. If this account is the backstop LP of the underlying collateral pool, it has some withdrawal windows. However, if the account doesn't belong to any collateral (that's the case until the account becomes active on any market), it cannot be backstop LP of that pool.

core/src/modules/AccountModule.sol
0xZenus: I used AccountExposure here instead of Account just to avoid duplicate code (I caught up with AB at some point and decided to migrate to this to remove redundant code).

core/src/modules/ExecutionModule.sol
0xZenus: because this condition is already checked in the marketActiveMarket function

core/src/modules/liquidation/CommonLiquidationModule.sol
0xZenus: same comment as above (I used AccountExposure here instead of Account just to avoid duplicate code (I caught up with AB at some point and decided to migrate to this to remove redundant code)).

integration/test/scenarios/core_scenarios/ScenarioA.sol
0xZenus: I'll make these vars explicit. I cannot bake them into the helper function because we will get rid of dated irs and I want that function to be generic.

0xZenus: 31a2c79a578aa65f88e49ef71edcdade99a5e408

ioanaroceanu2000 (approved):

PR #214, opened 2023-10-25
[DROP] Dated IRS instrument

0xZenus:


avniculae:
can you remove propagate ADL from the core as well?

avniculae (approved):

PR #216, opened 2023-10-26
feat: check requested withdrawal amount

avniculae:


core/src/libraries/account/AccountCollateral.sol
```diff
@@ -95,4 +133,51 @@ library AccountCollateral {
...
+            collateralPool.id, collateralType, address(0)
+        );
+
+        uint256 withdrawalAmountInUSD = getExchangedQuantity(
+            withdrawalAmount.toInt(), exchangeInfo.price, exchangeInfo.priceHaircut
+        ).toUint();
+
+        MarginInfo memory accountMarginInfo = self.getMarginInfoByBubble(address(0));
+
+        if (withdrawalAmountInUSD.toInt() > accountMarginInfo.initialDelta) {
```
arturbeg, line 173: hmm, is this not just equivalent to doing an IM check at the end of the execution module?

core/src/libraries/actions/EditCollateral.sol
```diff
@@ -157,6 +160,8 @@ library EditCollateral {
             }
         }
 
+        account.checkWithdrawalAmount(collateralType, tokenAmount);
```
arturbeg, line 163: alternatively could just do checks after updateNetCollateralDeposits? both the im check and real balance > 0 check? this could work as well tho, keen to hear @0xZenus thoughts on this as well

arturbeg (approved):

PR #224, opened 2023-10-29
Local LM Check 

arturbeg:


core/src/libraries/account/AccountAutoExchange.sol
```diff
@@ -259,11 +259,7 @@ library AccountAutoExchange {
         vars.quoteToCollateralExchangeInfo =
             CollateralConfiguration.getExchangeInfo(collateralPool.id, quoteToken, collateralToken);
 
-        // todo: need to double check the premium calculation implementation, consider renaming the discount to premium
-        // stETH -> USDC
-        // stETH -> ETH -> USD -> USDC
-
-        vars.autoExchangeBonus = UNIT.add(vars.quoteToCollateralExchangeInfo.autoExchangeDiscount);
+        vars.autoExchangeBonus = vars.quoteToCollateralExchangeInfo.autoExchangePremium;
```
ioanaroceanu2000, line 262: considering `collateralToLiquidate = vars.autoExchangeBonus * vars.priceQuoteToCollateral * amountToAutoExchangeQuote` I feel like `1 + percentage` is a better format for the bonus. i.e. when setting it in config, it makes more sense to set premium = 30% rather than 130%

core/src/storage/CollateralConfiguration.sol
```diff
@@ -321,6 +321,10 @@ library CollateralConfiguration {
         return UNIT.sub((UNIT.sub(a)).mul(UNIT.sub(b)));
     }
 
+    function combinePremiums(UD60x18 a, UD60x18 b) private pure returns (UD60x18) {
+        return UNIT.add((UNIT.add(a)).mul(UNIT.add(b)));
```
ioanaroceanu2000, line 325: can we add this function and the `combineDiscounts` to the price utils library?

core/src/storage/Account.sol
```diff
@@ -67,7 +67,7 @@ library Account {
     /**
      * @dev Thrown when a given account's margin balance is below the adl margin requirement
      */
-    error AccountBelowAdl(uint128 accountId, MarginInfo marginInfo);
+    error AccountBelowAdl(uint128 accountId, MarginInfo marginInfo, MarginInfo localMarginInfo);
```
ioanaroceanu2000, line 70: I don't have context over this localMarginInfo change, could you point me to the discussion or ask someone else to review this part?

arturbeg: sure, so the local margin info check means that a bubble can only be liquidatable if:
- address(0) bubble is liquidatable
- and token bubble is also liquidatable

core/src/libraries/account/AccountAutoExchange.sol
arturbeg: hmm, i think it is already set as percentage since we do UNIT.add() in combine premiums unless missing smth?

UNIT.add((UNIT.add(a)).mul(UNIT.add(b)));

integration/test/scenarios/core_scenarios/auto-exchange/ScenarioAE_A.sol
```diff
@@ -178,7 +178,7 @@ contract ScenarioAE_A is ScenarioFramework {
                     hasParent: true,
                     tokenAddress: address(0),
                     priceHaircut: ud(0),
-                    autoExchangeDiscount: ud(0),
+                    autoExchangePremium: ud(0),
```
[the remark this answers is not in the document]
arturbeg: why should it break the tests? see https://github.com/Voltz-Protocol/v2-core/pull/224#discussion_r1376011453

core/src/storage/CollateralConfiguration.sol
arturbeg: hmm, don't have a strong view on this, since these functions are only ever used in collateral configuration library on the surface it makes sense to keep them there as private?

PR #225, opened 2023-10-30
Feat/unified backstop adl

avniculae:


core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -613,190 +615,122 @@ library AccountLiquidation {
         // revert if account is not below adl margin requirement
         MarginInfo memory marginInfo = self.getMarginInfoByBubble(address(0));
         MarginInfo memory localMarginInfo = self.getMarginInfoByBubble(quoteToken);
-        if ((marginInfo.adlDelta > 0) && (localMarginInfo.adlDelta > 0)) {
+        if (marginInfo.adlDelta > 0 || localMarginInfo.adlDelta > 0) {
```
avniculae, line 618: @arturbeg note this change

arturbeg: good catch, however, we've just adjusted the flows in the latest PR and the actual condition for backstop liquidations now is:

```
        self.checkBelowGlobalLm();
        self.checkBelowLocalAdl(quoteToken);
```

this is because adl and actually also dutch deltas can only be local

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -613,190 +615,122 @@ library AccountLiquidation {
...
 
-        // adl maturities with positive upnl at market price
         uint256[] memory markets = self.activeMarketsPerQuoteToken[quoteToken].values();
+
+        if (markets.length != inputs.length) {
+            revert MismatchInputsMarketsLengths(self.id, quoteToken, inputs.length, markets.length);
+        }
+
+        // compute unrealized pnl volume
+        uint256 totalUPnLVolume = 0;
```
arturbeg, line 684: hmm, small point but the term PnLVolume feels hard to interpret?

core/src/libraries/account/AccountLiquidation.sol
```diff
@@ -613,190 +615,122 @@ library AccountLiquidation {
...
-            });
-        }
+            Market.Data storage market = Market.exists(marketId);
 
-        uint256 insuranceFundCoverAvailable =
-            collateralPool.getAvailableInsuranceFundCover({ quoteToken: quoteToken, preserveMinThreshold: true });
+            PnLComponents memory pnlComponents = market.getAccountPnLComponents(self.id);
 
-        MarginInfo memory quoteMarginInfo = self.getMarginInfoByBubble(quoteToken);
+            // upnl here is negative since all positive upnl exposures were adl-ed at market price
```
arturbeg, line 691: is this comment still true?

arturbeg (approved):

0xZenus (approved):

core/src/libraries/account/AccountLiquidation.sol
avniculae: addressed

core/src/libraries/account/AccountLiquidation.sol
avniculae: addressed

PR #226, opened 2023-10-30
feat: price with decimals

ioanaroceanu2000:


core/src/libraries/PriceHelpers.sol
```diff
@@ -9,13 +9,57 @@ pragma solidity >=0.8.19;
...
+    returns (uint256 amountOfTokenB)
+{
+    amountOfTokenB = convertAssets(amountOfTokenA.toInt(), decimalsDelta, price).toUint();
+}
+
+/**
+ * @dev When converting from token A to token B, use the following formula
+ * quantityTokenA * 10^(getDecimalsDelta(A, B)) = quantityTokenB
+ */
+function getDecimalsDelta(address fromToken, address toToken) view returns (int8) {
```
0xZenus, line 60: I'd avoid having view functions outside of libraries and so. Pure functions are fine to be here but I'd move this function to GlobalCollateralConfiguration.

core/src/libraries/PriceHelpers.sol
```diff
@@ -9,13 +9,57 @@ pragma solidity >=0.8.19;
...
+
+    return convertAssets(quantity, decimalsDelta, haircutPrice);
 }
 
 function getSign(int256 value) pure returns (int256) {
     return (value >= 0) ? int256(1) : -1;
 }
+
+function convertAssets(int256 amountOfTokenA, int8 decimalsDelta, UD60x18 price) pure returns (int256 amountOfTokenB) {
+    int256 amountOfTokenBInTokenADecimals = mulUDxInt(price, amountOfTokenA);
```
0xZenus, line 40: let's return in a single equation. I.e. `amountOfTokenB = DecimalMath.changeDecimals(mulUDxInt(price, amountOfTokenA), decimalsDelta)` -> this way we avoid tracking variable in audits and so.

0xZenus: As I see the code now, I'd actually have two more variables: tokenDecimalsA and tokenDecimalsB rather than delta.

core/src/libraries/account/AccountExposure.sol
```diff
@@ -75,47 +75,66 @@ library AccountExposure {
                 CollateralConfiguration.exists(collateralPool.id, tokens[i]);
             UD60x18 price = collateral.getParentPrice();
             UD60x18 haircut = collateral.parentConfig.priceHaircut;
+            int8 decimalsDelta = getDecimalsDelta(tokens[i], collateral.parentConfig.tokenAddress);
```
0xZenus, line 78: If we decide to pass tokenDecimalsA and tokenDecimalsB to the function, it's actually more efficient to get the token decimals of quoteToken out of the for-loop and get the decimals of tokens[i] here.

core/src/libraries/PriceHelpers.sol
```diff
@@ -9,13 +9,57 @@ pragma solidity >=0.8.19;
...
+import { DecimalMath } from "@voltz-protocol/util-contracts/src/helpers/DecimalMath.sol";
+import { SafeCastU256, SafeCastI256 } from "@voltz-protocol/util-contracts/src/helpers/SafeCast.sol";
 
-function getExchangedQuantity(int256 quantity, UD60x18 price, UD60x18 haircut) pure returns (int256) {
+import { GlobalCollateralConfiguration } from "../storage/GlobalCollateralConfiguration.sol";
+
+using SafeCastU256 for uint256;
+using SafeCastI256 for int256;
+
+function getExchangedQuantity(
```
0xZenus, line 20: Can we change the argument of this function? To int256 quantity and ConversionData struct that consists of price, haircut, tokenDecimalsA and tokenDecimalsB. And when we call this function >10 times, we can pass 2 arguments each function rather than 4-5 arguments each. 

Actually, it'd be great if we can use the same struct ExchangeInfo from CollateralConfiguration (and move the struct to DataTypes.sol). It'd be quite clean this way.

ioanaroceanu2000: Yes that's actually nicer. I removed the struct to DataTypes and used it as a param in [94969ed](https://github.com/Voltz-Protocol/v2-core/pull/226/commits/94969edcaece95812ac969f5f14524bdbcc207a4)

core/src/libraries/account/AccountExposure.sol
ioanaroceanu2000: I think it's cleaner to use the delta, we don't have to worry about the direction and carry two vars

core/src/libraries/PriceHelpers.sol
ioanaroceanu2000: I minimised the code to only return this equation `DecimalMath.changeDecimals(mulUDxInt(price, amountOfTokenA), decimalsDelta)` in [94969ed](https://github.com/Voltz-Protocol/v2-core/pull/226/commits/94969edcaece95812ac969f5f14524bdbcc207a4)

core/src/libraries/PriceHelpers.sol
ioanaroceanu2000: yes, changed in [94969ed](https://github.com/Voltz-Protocol/v2-core/pull/226/commits/94969edcaece95812ac969f5f14524bdbcc207a4)

